PR #97, opened 2015-02-04
Large update for the AncMed variant.

Sleepcap:
-> Some borderissues fixed
 -> new icons
 -> and no "L"-shape in smallmap

Signed-off-by: Oliver Auth <Oliver.Auth@rhoen.de>


TimothyJones:
Sorry it took so long to get back to you on this- I had a discussion with Tuan, and:

* we'd like to keep the ship/tank icons on the main site for consistency 
* and for now keep the variant the same as the variant page linked to from the description (with the convoy rule for balereas). 

Are you able to amend the pull request easily? If it's too much trouble, I can do it at our end.

Also, do you have a summary of the border fixes? With the white space changes its hard to see where the text has changed.

Thanks again for this fix!

Sleepcap:
Hi Tim.
What do you mean with "keep the variant the same as the variant page linked to from the description (with the convoy rule for balereas)". The ships where never able to convoy from the balereas in the variant. It's technically not easy to implement this, as the balereas are a "coast"-territory, and ships on  a coast can't help with a convoy...

TimothyJones:
Ah, apologies - I may have been confused. I thought this update included a rule change:

> 1.7:   new: rules.html added to explain the rule-change for the Baleares.

Are you saying that this was always the rule, but it was just never explained anywhere? If that's the case, we definitely want to update the description.

PR #107, opened 2015-02-17
Update for the AncMed variant.

Sleepcap:
* fixed a graphical error in the smallmap
* added a missing border on the maps
* rules.html added to show in the variant description that the fleets in the Baleares can't convoy.

Signed-off-by: Oliver Auth <Oliver.Auth@rhoen.de>


Sleepcap:
Did a new pull-request for the AncMed variant without the new icons.
Sorry, but I can't tell what border-error got fixed in the install.php. I use a tool to add these borders, and I didn't take notes about this.

Issue #126, opened 2015-04-16
Correct "Cancel game" description text


TimothyJones:
Both the admin CP and the TD powers "Cancel game" text read:

> "Splits points among all players in a game equally, and deletes the game."

This is incorrect. The points each player has bet are refunded, not an equal split of the pot. This should be corrected.

TimothyJones:
Fixed by the above pull request from humunuk

PR #128, opened 2015-04-16
Update adminActions.php Issue #126

humunuk:
Change description for issue #126

### DIFF
--- a/admin/adminActions.php
+++ b/admin/adminActions.php
@@ -38,7 +38,7 @@ class adminActions extends adminActionsForms
 			),
 			'cancelGame' => array(
 				'name' => 'Cancel game',
-				'description' => 'Splits points among all players in a game equally, and deletes the game.',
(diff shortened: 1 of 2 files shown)


TimothyJones:
Thanks! Can you make the same change in `adminActionsTD.php` too please, and then I'll accept this change. That way it's one pull request, not two.

humunuk:
I think it's done, let me know if something wrong, learning git(Hub).

TimothyJones:
Great, thanks! I'll have a quick test now, and then accept the change. Normally the description would also include the testing that you've done (ie, in this case verified that the pages still look correct in the browser), but since this is a very easy test, I'll do it myself.

I've just updated the [developer instructions](http://webdiplomacy.net/developers.php) on the website to reflect this. It's only just been updated though, so there's no way you'd have seen this before you made the pull request :)

TimothyJones:
Thanks for the change! Much appreciated!

PR #141, opened 2015-05-10
A mechanism for user options, and client side colourblind fix

TimothyJones:
Adds `objects/useroptions.php`, an object for handling user options. User options are stored in their own table, which is only populated when users save options (so it is lazily created for existing members). They are loaded when a `User` object is created. In addition to being part of the user object, options are also exposed to the client through `useroptions.php`, which produces a json object that describes a user's preferences for use in javascript.

The colorblindness fix is applied client side, depending on the user's options. It is triggered when the map changes, and replaces the map image with an appropriately adjusted version. Additionally, press is displayed with the country's name at the start of the message. Thanks to @Sleepcap for the suggestions and inspiration of this feature.

Additionally, the options contain an option to always hide the move arrows on the board, and to hide the new live game announcement on the home page.

Feedback welcome.

### DIFF
--- a/install/1.37-1.38/update.sql
+++ b/install/1.37-1.38/update.sql
@@ -1,9 +1,11 @@
-CREATE TABLE `wD_WatchedGames` (
-		  `userID` mediumint(8) unsigned NOT NULL,		
-		  `gameID` mediumint(8) unsigned NOT NULL,		
-		  KEY `gid` (`gameID`),		
-		  KEY `uid` (`userID`)		
-	) ENGINE=InnoDB DEFAULT CHARSET=utf8;		
-		
-		
--UPDATE `wD_Misc` SET `value` = '138' WHERE `name` = 'Version';
+ CREATE TABLE `wD_WatchedGames` (
+	  `userID` mediumint(8) unsigned NOT NULL,
+	  `gameID` mediumint(8) unsigned NOT NULL,
+	  KEY `gid` (`gameID`),
+	  KEY `uid` (`userID`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8;
+
+ALTER TABLE wD_Games ADD `minimumReliabilityRating` tinyint(3) unsigned NOT NULL DEFAULT '0';
+ALTER TABLE wD_Backup_Games ADD `minimumReliabilityRating` tinyint(3) unsigned NOT NULL DEFAULT '0';
+
+UPDATE `wD_Misc` SET `value` = '138' WHERE `name` = 'Version';


board/chatbox.php
```diff
@@ -421,6 +421,15 @@ public function renderMessages($msgCountryID, $messages)
 			if( is_object($Member) && $message['fromCountryID'] == $Member->countryID )
 				$message['message'] = '<span class="messageFromMe">'.$message['message'].'</span>';
 
+			// Display the country name in front of the text (for colorblind people)
+			if ( $User->options->value['colourblind'] != 'Yes')
```
thorfi: if ( $User->options->value['colourblind'] != 'No') // ?

objects/useroptions.php
```diff
@@ -0,0 +1,129 @@
...
+		$update = implode(',',$updates);		
+
+		$row = $DB->sql_hash("SELECT * FROM wD_UserOptions WHERE userID=".$this->id );
+		if ( ! isset($row['userID']) or ! $row['userID'] ) {
+			// create
+			$DB->sql_put("INSERT INTO wD_UserOptions (userID) VALUES (".$this->id.")");
+		} 
+		
+		$DB->sql_put("UPDATE wD_UserOptions SET $update WHERE userID=".$this->id);
+	}
```
thorfi: Is $DB auto committing, is there a commit somewhere else, or should there be one here?

locales/English/user.php
```diff
@@ -86,7 +86,19 @@
...
-<?php
+
+<?php 
+	foreach ($User->options->value as $name=>$val) 
+	{
+		print '<li class="formlisttitle">'.UserOptions::$titles[$name].':</li>';
+		print '<li class="formlistdesc">';
+		foreach (UserOptions::$possibleValues[$name] as $possible) 
+		{
+			print ' <input type="radio" name="userForm['.$name.']" value="'.$possible.'" '. ($val == $possible ? 'checked' :'') . ' > '. $possible;
```
thorfi: Not sure if you want to have a title="" element to explain the meaning of values (particularly the colour blind ones)?

javascript/mapUI.js
```diff
@@ -137,11 +137,23 @@ function loadMap(gameID, currentTurn, newTurn)
 	
 	// Update the link to the large map
 	$('LargeMapLink').innerHTML = 
-			' <a href="map.php?gameID='+gameID+'&turn='+newTurn+'&mapType=large" target="blank" class="light">'+
+			' <a href="map.php?gameID='+gameID+'&turn='+newTurn+'&mapType=large'+(useroptions.showMoves =='No'?'&hideMoves':'')+'" target="blank" class="light">'+
 			'<img src="'+l_s('images/historyicons/external.png')+'" alt="'+l_t('Open large map')+'" ' +
 			'title="'+l_t('This button will open the large map in a new window. The large ' +
 			'map shows all the moves, and is useful when the small map isn\'t clear enough.')+'" /><\/a>';
```
thorfi: title text might need fixing for the large map if you are hiding moves on the large map also?

board/chatbox.php
TimothyJones: Good catch. Fixed.

objects/useroptions.php
kestasjk: Data is committed once the page has finished being generated (assuming no errors occur), but note that MyISAM tables aren't covered by transactions and they auto commit, only the innodb tables are protected by transactions.

Issue #62, opened 2015-01-03
Credits page Needs Auto Mod list


jmo1121109:
Have the list shown in the admin cp under status lists also show up on the credits page so that devs don't have to update core files.

TimothyJones:
Fixed!

PR #167, opened 2015-08-22
Fix for #62: Automated moderators list for Credits page

jkleban:
First submitted pull request for fixing issues!  Let me know if there is official testing I can/should submit.

### DIFF
--- a/locales/English/credits.php
+++ b/locales/English/credits.php
@@ -27,6 +27,18 @@ defined('IN_CODE') or die('This script can not be run by itself.');
 ?>
 <div>
 <?php
+
+// Get moderator list
+global $DB;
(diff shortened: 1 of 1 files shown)


TimothyJones:
The "official testing procedure" such as it exists is to describe the testing as part of the pull request [for example, here](https://github.com/kestasjk/webDiplomacy/pull/38). This prevents me having to repeat the same tests that you've done :)

The testing process is the part that takes the most time - I typically spend more time testing than I do coding.

TimothyJones:
Thanks! Looks like this needs spacing between the names, too:

screen shot 2015-08-25 at 9 35 38 pm

Issue #149, opened 2015-05-19
Put scoring system on the game overview on the home page


TimothyJones:
Currently the scoring system (WTA or PPSC) is only listed on the main game page. This should be listed on the home page small game summary too.

PR #168, opened 2015-08-22
Fix for #149: Expose pot type in small gamepanel view on home page

jkleban:
Decided to just use a switch statement for generating the abbreviations for the pot type rather than make database updates.

### DIFF
--- a/gamepanel/gamehome.php
+++ b/gamepanel/gamehome.php
@@ -150,7 +150,23 @@ class panelGameHome extends panelGameBoard
 	 */
 	function pot()
 	{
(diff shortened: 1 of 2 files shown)


TimothyJones:
I'm not sure it makes the most sense for the game object to know what the abbreviations are - it feels like this is better done at presentation time, ie in `titleBar()` in `gamepanel/gameHome.php`. Is there a good reason to have the abbreviations on the game object instead?

TimothyJones:
I really like how this looks - having the pot type next to the pot is much nicer than the obvious approach of having it in the list of other game options.

Issue #132, opened 2015-04-21
Make it more obvious that clicking each tab on the "Games" page brings up the search


TimothyJones:
Currently it's not clear that tabs on the Games page can be clicked to search within that tab. A small interface change (eg, putting a little down arrow on each tab) would make this clearer.

PR #170, opened 2015-08-23
Fix for #132: advanced game search capability highlighted on active g…

jkleban:
…amelistings tab


TimothyJones:
This is great! I'm just heading out, but I'll accept this tonight.

It will also need an entry in `locales/English/recentchanges.php`, but I can add one when I accept it if you haven't done so by then.

jkleban:
I'll let you add the comment to recentchanges and date it exactly when the pull request is accepted.

Issue #201, opened 2015-12-28
Show global press to all users even if they are not in the game.


TimothyJones:
When viewing the game panel, Moderators can see the global press above the game even if they are not in it. Non-moderators have to click through to the messages archive.

There's no reason that regular users shouldn't be able to see the global chat panel above the game.

deltachemist:
I've started working on this issue.

PR #214, opened 2016-01-14
Bugfix #201

deltachemist:
The chat panel is now visible to every visitor of the game, once the game started.

**Test cases, checked and passed**
* Tested under Firefox
* Created a game with no chat restrictions.
** Mods/Admins see the global tab and are able to write messages.
** Participants see all tabs and are able to write in every tab.
** Spectators (or Observers, they don't have to select "Spectate") can see the global tab but nothing else. They are not able to write in the global tab.

* Created a game with no ingame-messaging
** Mods/Admins are able to see the global tab and write in it.
** Participants are able to see the global and notes tab, but can only write in the notes tab.
** Spectators are able to see the global tab

**Test cases not checked**
* A game with global press only

**Possibly unexpected/unwanted behaviour**
* In games with no chat, spectators will still be able to see the Global Tab (although it will mostly be empty)
* Users that are not logged on are also able to see every chat panel as well
* Before the game starts no one is able to see the chatbox (might be confusing for users that are only spectating?)

### DIFF
--- a/board.php
+++ b/board.php
@@ -246,7 +246,6 @@ if( isset($Member) && $Member->status == 'Playing' && $Game->phase!='Finished' )
 	}
 }
 
-//if ( 'Pre-game' != $Game->phase && ( isset($Member) || $User->type['Moderator'] ) )
 if ( 'Pre-game' != $Game->phase )
 {
 	$CB = $Game->Variant->Chatbox();
--- a/board.php
+++ b/board.php
@@ -246,7 +246,8 @@ if( isset($Member) && $Member->status == 'Playing' && $Game->phase!='Finished' )
 	}
 }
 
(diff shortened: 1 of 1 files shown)


TimothyJones:
Thanks very much for this! Sorry it took me a while to get to it - I'll confirm your testing when I get a change and accept this change.

Could you be more specific about what you mean by this line:

> Users that are not logged on are also able to see every chat panel as well

Does that mean that they see *all* country tabs? Or just global. Just global is desirable, other country tabs less so (although I suspect they won't be able to see press even if the tabs are there).

deltachemist:
> Does that mean that they see all country tabs?

No, users that are not in the game (regardless if they are logged on or not) can only see the global panel.

PR #229, opened 2016-04-16
Display message when leaving page with unsaved orders

deltachemist:
Tested under Opera and Firefox

* Opera and Firefox
   * Leaving any page with unsubmitted text will still trigger the warning
   * Leaving the board with unsaved orders will trigger a warning about unsaved orders
   * In case the user selects to leave the page with both unsaved orders and unsubmitted text only one warning (in my case unsaved orders but this does not seem to be guaranteed, according to documentation of addEventListener) will be triggered (even if the user decides to leave the page)

* Special case Firefox
   * No custom text is displayed leaving the user without knowledge why the page requests the navigation

Untested browsers: Internet Explorer (according to specifications of addEventListener some IE versions need to call attachEvent), mobile browsers (according to specifications of unbeforeunload it seems that some mobile browsers seem to (or have been) ignoring the unbeforeunload events.

### DIFF
--- a/javascript/orders/form.js
+++ b/javascript/orders/form.js
@@ -177,11 +177,13 @@ function OrdersHTMLFormClass() {
 			if( MyOrders.pluck('isChanged').any(function(c){return c;}) )
 			{
 				this.ordersChanged=true;
+				window.ordersChanged = true;
 				this.buttonOn('UpdateButton');
 			}
 			else
 			{
 				this.ordersChanged=false;
+				window.ordersChanged=false;
 				this.buttonOff('UpdateButton');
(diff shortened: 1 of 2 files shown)


TimothyJones:
Sorry for the delay on this! I'll check it out and accept it next week.

TimothyJones:
Done! If you remind me your webdip username, I'll make a forum post crediting you.

deltachemist:
That would be doktorkeks42.

Am 3. Mai 2016 06:16:58 MESZ, schrieb Timothy Jones <notifications@github.com>:

PR #230, opened 2016-04-16
Conversion of (D) into points symbol

deltachemist:
Changed the regex for detection of the points icon to only detect "(d)" [was "d" and "points"].

Attached image shows new outputs. The row with the symbols have been "1(D)", "1(d)", "1 (D)" and "1 (d)"

points


TimothyJones:
Sorry for the delay on this! I'll check it out and accept it next week.

PR #233, opened 2016-04-30
Install and update utilities

deltachemist:
Added windows (batch) and linux (bash) script that create a install_full.sql file with all updates. If anyone creates a new database update he can run one of the two files to create the new install_full.sql file. Anyone who wants a new database install can then simply execute install_full.sql and installs the basic install and all updates.

Also added update.php (and linked it from install.php) in order to automatically update the database if it's version is behind the code version. This requires the user to be an admin.

Both parts require a certain directory structure. The actually update file must be named update.sql and be located in a folder called previousversion.previoussubversion-newversion.newsubversion, where previousversion and newversion are postive integers and previoussubversion and newsubversion are positive two-digit integers. Also version numbers need to be increasing (seems selfevident but is also required). At the moment there is no sorting of the folders containing the update files. This might be a problem in the future but at the moment everything works as intended.

### DIFF
--- a/install/create_installscript_linux
+++ b/install/create_installscript_linux
@@ -1,0 +1,8 @@
+#!/bin/sh
+#When editing this file please make sure to also update create_installscript_windows.cmd
+cp install.sql install_full.sql
+for VAR in $(find . -name 'update.sql' | sort); do
+   echo Appending $VAR
+   cat $VAR >> install_full.sql
+done
+echo Installscript successfully generated.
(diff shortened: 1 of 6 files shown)


TimothyJones:
Hi, thanks for this! I'm concerned that there might be problems if two admins load the site at the same time after a database update - causing the update script to double-bounce. What do you think?

TimothyJones:
I'm going to close this for now as https://github.com/TimothyJones/webDiplomacy-dev-docker provides much of the same functionality. Feel free to reopen if you think we should still merge it.

Issue #218, opened 2016-02-01
URL section on profile page should be truncated


TimothyJones:
If the URL/homepage field on a user's profile is very long, then the text overlaps with the statistics section:

![screen shot 2016-02-01 at 2 36 46 pm](https://cloud.githubusercontent.com/assets/914369/12708216/4781235e-c8f1-11e5-8d30-3174f90dc266.png)

Instead, the text should be truncated or wrap.

yewang:
Addressed by merged PR https://github.com/kestasjk/webDiplomacy/pull/239

jmo1121109:
This is fixed, closing out.

PR #239, opened 2016-06-07
Fix Issue #218

yewang:
Added word-wrap:break-word style to list item to wrap long home page URLs

### DIFF
--- a/profile.php
+++ b/profile.php
@@ -524,7 +524,7 @@ if ( $UserProfile->hideEmail != 'No' )
 
 if ( $UserProfile->homepage )
 {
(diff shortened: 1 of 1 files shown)


TimothyJones:
Thanks! How would you like to be credited? Can you add a change for the recentchanges.php file too, please?

yewang:
What are the conventions for the dates on the recent changes file? Is it the date that the code was merged or when the change was deployed to server? I can write the text snippets for this and the other PR.

"Long home page URLs on profile pages now wrap."

As for credit, these are just some very minor contributions, but if you want, you can list "Minor fixes" credited to my WebDip account name (yebellz, userID=738).

PR #243, opened 2016-07-28
Move password box out of advanced settings

raj-kesavan:


TimothyJones:
Thanks for the change!

Issue #228, opened 2016-03-19
CSS: Chat text doesn't wrap


LitterPicker:
A player sent me a chat message reading "With what!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!?!?!??!?!?!?!?!?!!!!?!!!??!?!??!!?!?!??!?!?!?!?!??!?!?!?!??!?!?!!??!?!?!?!?!!!????!?!?!??!"
There were undoubtedly a few more !s and ?s after that, but the end of the message cannot be read, and neither, now, can the ends of long lines in other messages.
My response, for example:
"Do you want to keep playing? I'll give you your country back if you like. There would be some conditions, of co"
That's not quite what I wrote.

TimothyJones:
This is probably just a CSS change. What browser are you using? My browser (Chrome) adds a horizontal scroll bar when this happens.

Either way, I believe a workaround is to highlight the text and copy/paste it out.

LitterPicker:
I generally use Firefox, but I checked the chat in Chrome, and I don't see a horizontal scroll bar.

jmo1121109:
Tested this out in depth, can no longer replicate and looks corrected in the css. Closing this out.

PR #267, opened 2017-11-16
Better mobile support

Zjcompt:
This is a build I made to better support mobile devices. I was kinda sick and tired of having to zoom in constantly and how most things were very hard to do on a mobile device. This branch adds full mobile support, focused on CSS changes only. There were a few non css files that I had to edit because of inline styling issues.

Everything is working properly in all major browsers:

Tested in:
Edge & IE11
Chrome 49
Safari 10
Firefox 48

No github issue, just want to make the site/project better!

Let me know if there are any concerns or questions!

### DIFF
--- a/board/info/messages.php
+++ b/board/info/messages.php
@@ -41,7 +41,7 @@ print '<h4>'.l_t('Chat archive').'</h4>';
 print '<div class="variant'.$Game->Variant->name.'">';
 
 $CB = $Game->Variant->Chatbox();
-print '<table>'.$CB->getMessages( -1, $pager->SQLLimit()).'</table>';
+print '<table class="archive-messages-table">'.$CB->getMessages( -1, $pager->SQLLimit()).'</table>';
 
 print '</div>';
 
--- a/css/gamepanel.css
+++ b/css/gamepanel.css
@@ -495,4 +495,7 @@ div.memberBoardHeader {
 	.orderEnd{
 		display: none;
 	}
+	#mapImage{
+		width: 100%;
+	}
 }
--- a/css/gamepanel.css
+++ b/css/gamepanel.css
@@ -412,3 +412,87 @@ div.memberBoardHeader {
 	border-top: 1px solid #cecece!important;
 	padding:0 5%;
 }
+
+@media only screen and (max-width: 800px) {
+	.titleBarRightSide {
+		float: none;
+		text-align: center;
+	}
+	.titleBarLeftSide {
+		float: none;
+		text-align: center;
+	}
+	.gamePanelHome{
+		font-size: 16px;
+	}
+	.gamePanelHome .titleBar{
+		font-size: 14px;
+	}
+	.gamePanelHome .homeMembersTable td {
+		display: table-cell;
+	}
+	.gamePanelHome .homeMembersTable tr {
+		display: table-row;
+	}
+	.gamePanelHome .homeMembersTable td {
+		display: table-cell;
+	}
+	.gamePanelHome .homeMembersTable tr {
+		display: table-row;
+	}
+	.gamePanelHome .occupationBarTable td {
+		display: table-cell;
+	}
+	.gamePanelHome .occupationBarTable tr {
+		display: table-row;
+	}
+	.gamePanelHome .memberBoardHeader td {
+		display: table-cell;
+	}
+	.gamePanelHome .memberBoardHeader tr {
+		display: table-row;
+	}
+	.gamePanel {
+		width:100%;
+	}
+	.homeGameLinks{
+		font-size: 16px;
+	}
+	.memberUserDetail {
+		float:none;
+		text-align: right;
+		overflow: hidden;
+	}
+	.memberUserDetail > input{
+		margin-top: 10px;
+	}
+	.archiveBar{
+		text-align: center;
+	}
+	.memberName{
+		white-space: normal;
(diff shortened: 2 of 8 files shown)


TimothyJones:
This is amazing! Thank you so much.

I'll check it out as soon as I can - might not be for a couple of weekends.

TimothyJones:
Also, let me know your webdip user name and I'll make sure you get the proper credit.

Can you tell me what mobile devices/browsers you tested on?

Sleepcap:
With your patch there chat no longer makes a linebreak at the end of a word, but just mid-word...

Zjcompt:
Timothy,

Username (by accident) is: Exectuor

Tested on Android/iOS and Desktop. Specific browsers is on the OP.

--------------------------------

Sleepcap,

I believe I did this change to specifically fix issues like this one:

https://github.com/kestasjk/webDiplomacy/issues/228

What happens is with words that don't have a breaking space, they just go WAYYYYY off the side of the page. It looks a lot worse to me personally. The compromise is that words will break. Do words break on both desktop and mobile? It is more important on mobile because of how much smaller the chat box is.

Sleepcap:
But now it splits words even if there are no really long words wight bevore the last letter.
I'm not sure if it's better or worse...

Zjcompt:
Right, like I said this is the compromise that words will break, I may be able to add a hyphen at the break. I will try to see if I can get a better solution included.

Zjcompt:
@Sleepcap The new commit should fix the issues you were talking about.

Sleepcap:
Works great now.
I will use this on vDiplomacy, so we can see if something important breaks...

Sleepcap:
It's live on vDip. My users already started discussing. There are a few ideas for improvement already.
http://www.vdiplomacy.com/forum.php?viewthread=75175

Zjcompt:
I put another update in so that My Games comes first. I agreed with your users on that. Sometimes you have to scroll really far to get to them.

Zjcompt:
I also noticed that sometimes you use an iframe for the game image. I added a fix so that it doesn't go way off the side of the screen. As for your users that kinda don't like the mobile version you can send them these three links to go to the old non mobile version:

https://ios.gadgethacks.com/how-to/request-desktop-sites-ios-9s-safari-for-ipad-iphone-0163057/
https://www.cnet.com/how-to/how-to-request-desktop-version-of-a-web-site-in-chrome-for-android/
https://support.mozilla.org/en-US/kb/how-do-i-change-desktop-view-firefox-android

Sleepcap:
Thanks, works great so far. The few people complain so far should be able to get the desktop site without any problems. If they really need a setting to "force" the desktop-site I will move the CSS code in an additional file and exclude the file on demand.
Another unrelated question (I have no CSS knowledge, so this is really hard for me): I have a background image for the site. It has a fixed position on all desktop-browsers, but not on iOS. Also maybe it should not load at all on mobile. There is not much too see of it on such small devices.

In the body-CSS I use:
	background:url(../../images/background.jpg) center top repeat-y #dcbea4;
	background-attachment: fixed;

TimothyJones:
Merged and pushed to the site. Thank you so much for this change!

TimothyJones:
@Sleepcap - Do you think you could produce a PR with the opt out fix? I tried to cherry-pick it from the vDip commit history, but it didn't turn out to be straightforward. I'd be very grateful for the contribution!

Sleepcap:
Unfortunately it's really not 100% straightforward, as I have 2 different CSS-styles (webdip and vDip color-scheme), and I moved all my own options to another settings page. Changing this back to the webdip-style needs some time, and I'm currently porting some other variants to the site. So this would take some time.
Basically you just need to put the new mobile-code in a separate CSS-file and only include this in the lib.html if the user has _not_ opt-out in his setting. So if the user has activated the opt-out the mobile-css would not load at all.

kestasjk:
Great work all, and many thanks Zach. Tim are people not okay with changing
their browsers to desktop mode? Mobile browsers almost always have an
option to choose modes for just this sort of reason.

Rgds,
Kestas

TimothyJones:
I think it’s done as a responsive site, rather than a mobile site, so request desktop site doesn’t change it

kestasjk:
Ah I see. What don't people like about it? Is this just aversion to change
or are there things that need to be improved?

TimothyJones:
One person said they use webdip on a desktop in two split screens, both less than 800 pixels wide, triggering the responsive site (one page for orders and one for the map). Probably not a common case. Another said they don’t like the larger font size on mobile. 

Although both of those issues are unlikely to be common, I still think it’s worth having an opt out.

kestasjk:
It looks like the preferred way of having an opt-out for responsive
websites is to put all the @media CSS into a different file, which can then
be included / excluded depending on a user option. Zach / Oliver since you
are more comfortable with the CSS structure could you do this? Once it's in
its own file I can take care of adding a column and user option etc to
allow it to be disabled.

Sleepcap:
Wanted to do a pull-request, but your current branch is not properly maintained at the moment.
in definitions.html the version is set to 143, but the install-files are only up to 142, so I can't test against your latest release.

TimothyJones:
Sorry about that- that’s my fault. There is no 143- you can safely drop that to 142. I’ll make that change in an hour or so. Then you can test with no conflicts.

PR #272, opened 2017-12-17
Mobile desktop toggle

Zjcompt:
This adds a client side toggle between desktop and mobile css. 

Unfortunately this solution means that CSS files will have to be duplicated between desktop and mobile supported versions. May not be the best solution but this will solve the problems some people have with the new mobile layouts. It will also allow you to have it turned off on your home pc and turned on on your phone.


TimothyJones:
You're an absolute legend. I'll push this to the site at lunchtime.

PR #273, opened 2017-12-18
Mobile desktop toggle Attempt #2

Zjcompt:
Sorry for the multiple pull requests! We may have pulled the trigger to early on it! LOL

No big deal though! So what I did here is I added back the old CSS from before the mobile changes and then made sure to remove and then re add the viewport meta tag when toggling desktop/mobile modes. This should give back a full legacy experience to mobile users. 

As for people having issues with their cache, they should see it pop up in a 1-3 days. Not much you can do when browsers hold onto old JavaScript and/or CSS files that long. You may be able to add an auto versioning to the end of the JS/CSS files. Like how global.css has ?ver=1 at the end, but that might be for another day.


TimothyJones:
No need to apologise- I’d always rather have multiple requests over none!

I’ll take a look over this as soon as I can. Maybe I can add the versioning too, so people don’t have to refresh. The mechanism should work for all CSS/JS files.

Thanks again!

PR #274, opened 2017-12-19
Mobile Desktop Toggle Attempt #3

Zjcompt:
Soooooo. Back at it again. I am dedicated to making everyone happy here. 

So because I only tested this on my local machine the load times were pretty much instantaneous and I never noticed the weird flashing of the css changing.

This fix basically moves the javascript that handled the changing of the viewport and css files and is now render blocking so the changes happen before content is displayed to the users. 

Also I did update the $jsVersion to 7. This however should probably be made a global var in the config. Rather than sitting as a local var in the html.php.

If you have any questions let me know.


Zjcompt:
WARNING: Don't take this yet! Same thing happened. I had to put my profiler to load at 3G speeds but the same thing happens. I have a fix in the works, but I need to do more use cases first. I'll see if I have it tomorrow.

Zjcompt:
This now working. Feel free to test and merge.

This will now keep the screen zoomed out after going to new pages AND you wont see any flashes on pages in either mode. What you may see instead is the new CSS VERY briefly and then switch to the old CSS.

I tested this on a simulated 500kb/s and 500ms ping connection.

TimothyJones:
Awesome! Thanks so much. I've pushed it to the site.

Issue #245, opened 2016-11-16
Remove country selection weight


TimothyJones:
As per #235, the country selection weighter does not work correctly. Perhaps we should just remove it and make it random again.

rjmcf:
The country selection problem is a version of the Assignment Problem, which has an optimal solution in the Hungarian Algorithm. I will implement this and see how long it takes to assign countries, if it's anything more than a few seconds then I'll just go about making selection random again.

TimothyJones:
This is great! Thanks!

It’s worth noting that the country selection weighting is also currently broken (the previous weights are not saved correctly because an update statement should be a create or update). I’m on mobile so it’s a bit of a pain to pull up the issue number- but I’ll link it when I get to a computer.

rjmcf:
No worries! Is it the one that you linked above? I'll have a look at that too while I'm here.

TimothyJones:
Aha! Yes, it is. Thanks very much!

rjmcf:
I've implemented the algorithm in Python (as I'm more familiar with it and it's easier to prototype in), and even with 10x10 (I think the largest variant type we have?) the solution is basically instantaneous, so this looks very viable. I'll get to work implementing it in PHP and then see if I can slot it in to the codebase.

TimothyJones:
Awesome! Looking forward to seeing what you come up with.

rjmcf:
Right, I now have a php class implementing the algorithm, which I've tested with a parser, so I know it can do what it needs to do. It's pretty fast to do the 10x10 problem, like less than a second. There are a few issues to deal with now I think. 

0). What's the best way to test this on a server like the one it'll be operating on? I don't want to take up precious time from other important procedures.
1). I need to make sure the coding style, comments etc is aligned with the codebase. That hopefully shouldn't take too long.
2). I'll need to look at how to slot it in where it's needed. I have some ideas for how to do that, but it's just something else that needs doing.

Cheers!

rjmcf:
Ok I've edited the coding style so it's hopefully more in line, I've also edited pregame.php locally so it should be ready to go. Before I make a pull request I would like to try and test it in place, but I'm not sure what's best. How do you guys test new code before? Have you set up your own webservers and databases at home?

TimothyJones:
Yes, usually we've just set up a dev machine. It's pretty straightforward, but I've written a harness to do this for you in docker, that you can find here: https://github.com/TimothyJones/webDiplomacy-dev-docker

[...]

Issue #235, opened 2016-05-03
Country Selection not Random?


Domphalos:
In the American Empire  map i have had Texas all 4 times I have played.  The probability of this occurring randomly is 1 in 10,000.  I had a similar issue with Austria on the Classic map.  The first 4 times I played a no chat version I got Austria all 4 times.  This is not as improbable as the previous example but still pretty improbable at 1 in 2303.

Domphalos:
Update.  It is now 5 times in a row for Texas, a 1 in 100,000 chance.

kestasjk:
Hmm, thanks for pointing this out.. I think there actually is an issue with country selection..

The way it works is whenever you get a country your weighting for that country is halved, and your chance of being that country next time is relative to your weighting for that country. But the system has to decide who to choose a country for first, since the second person won't be able to get that country.

The system is supposed to select the most unbalanced player first, since they deserve to get the country according to their unbalanced weightings. It does this by taking the standard deviation of the country weightings, and the player with the largest standard deviation is taken as the most unbalanced, and they get the first pick of country.


Having done a few scenarios in a spreadsheet though this isn't nearly as good a measure of who is the most unbalanced as I thought.. If you start balanced and get the same country 10 times in a row you have a standard deviation of ~0.1, but if you start balanced and get five unique countries, followed by those five same countries again you get a standard deviation of ~1.4.

[https://github.com/kestasjk/webDiplomacy/blob/57410df26f7681e90882ba1ab83c1c35537449fe/gamemaster/adjudicator/pregame.php#L94](url)

I think if the standard deviation code was replaced with an order based on the player who has a country with the lowest weighting it would give better results:
`foreach($chanceGrid as $userID=>$chances) 
{ 
	$chanceGrid[$userID] = ($chances = $this->balanceChances($chances)); 

	$minChance = 1.0;
	foreach($chances as $chance) 
		$minChance = min($minChance, $chance);

	$selectionOrder[] = $userID; 
	$standardDevs[] = 1.0 - $minChance; 
} `
(Of course the comments and name of the $standardDevs array would have to be changed for clarity)

TimothyJones:
There's another issue with the country selector, which is that it doesn't create entries in the weighting table for new players (so every time a new player starts a game, their country weightings are even). This has been a known issue for a while, affecting all new accounts since around 2013. I haven't gotten around to making an issue for it yet.

TimothyJones:
[...]

PR #281, opened 2018-01-20
Implement an optimal solution for country selection

rjmcf:
#245 #235 (these issues depend on each other)
Previously country selection has been random, or attempted to use the standard deviation of the chances of a player getting a particular country. This change instead uses the [Hungarian algorithm](https://en.wikipedia.org/wiki/Hungarian_algorithm) to solve the assignment problem optimally. It also solves the problem where entries weren't being saved in the wD_VariantData table, so now the play counts will affect the country selection.

- assignmentSolver.php contains the implementation of the Hungarian algorithm, which assigns the elements of one set to the elements of another, minimising or maximising the weights as desired. 

- variants/variantData.php has been changed to allow upsertion of int and float values to the table. This allows us to insert a row if it doesn't exist and update it if it does.

- pregame.php has been updated to use the assignment solver and to store updated play counts in the wD_VariantData table.

**Testing:**

- Extensive testing has been done on the assignmentSolver code to ensure correct operation, including generating problems [here](http://www.hungarianalgorithm.com/solve.php) up to 10x10 and comparing solutions. 

- In the development environment I have tested that countries are assigned, that assignments are different when making a new game with the same players, that entries are added to the wD_VariantData table, and that entries are updated there too.

If more comments are needed in the code then I am happy to add them. Unfortunately git has decided that I deleted the entire pregame.php file and rewrote it, which is annoying. I'm not sure if there is a better way to do this to make sure that doesn't happen.


TimothyJones:
Thanks so much for this! I'll carve out some time to do a review over the weekend.

One note - you say that performance at 10x10 is ok - but the largest map that webdiplomacy ships with is 17x17. Does it still perform at that speed?

TimothyJones:
I think it thinks you've re-written that file because the permissions have changed for some reason.

rjmcf:
Thanks for having a look over this. The issue with testing a 17x17 configuration is that there's nothing I can check it against. With up to 10x10 I can generate and check problems with that website I linked, but above that there’s no way that I can find of making sure I’ve got the right answer. I could potentially do it by hand, but I’d rather not.

Also before I start trying to test it I’d like to know if there’s an easy way to set up a game with a number of new players without me having to individually register the players and join as them. This was bad enough testing 7x7 maps, any more and I might lose it. If this doesn’t exist already, this might be a nice developer tool to work on.

Different permissions will be my fault, I hadn’t realised that sort of thing would be an issue. There was a time when I had two versions of the file, pregameOld.php and pregameNew.php, one of which I would rename to pregame.php depending on which version of the assignment I wanted to use. Sorry about that, I’ll be more wary of it next time.

Cheers,
Rjmcf

On 23 Jan 2018, at 02:11, Timothy Jones <notifications@github.com<mailto:notifications@github.com>> wrote:


I think it thinks you've re-written that file because the permissions have changed for some reason.

rjmcf:
I just ran a test with a World map, 17 countries, and there was no noticeable extended delay during the assignment, it seemed to load as fast as a standard map has been.

On 23 Jan 2018, at 02:11, Timothy Jones <notifications@github.com<mailto:notifications@github.com>> wrote:


I think it thinks you've re-written that file because the permissions have changed for some reason.

gamemaster/adjudicator/pregame.php
```diff
@@ -1,322 +1,231 @@
...
+		
+		$countryCount = count($Game->Variant->countries);
+		for($i = 0; $i < count($userIDs); $i++)
+		{
+			$userCountryCounts[$i] = array();
+			$vd->userID = $userIDs[$i];
+			
+			/**
+			 * Note that the first country ID is 1 not 0.
+			 * assignmentSolver requires that the indexes start at 0, so we need to
```
TimothyJones: It feels like `assignmentSolver` should do this conversion, rather than expecting it to be done before the call. This way the contract for `assignmentSolver` is written in the same data schema as the rest of the code.

gamemaster/adjudicator/pregame.php
```diff
@@ -1,322 +1,231 @@
...
-		$this->assignCountries($userCountries);
-
-		// Create starting board conditions, typically based on $countryUnits
-		$this->assignTerritories(); // TerrStatus
-		$this->assignUnits(); // Units
-		$this->assignUnitOccupations(); // TerrStatus occupyingUnitID
-	}
-}
-
+<?php
```
TimothyJones: The permissions on this file have changed - we should probably leave them as is, unless there's a good reason to change.

gamemaster/adjudicator/pregame.php
rjmcf: Is it possible that countryId might one day not start at 1? Or some weird variant might have a particular way of assigning country IDs? That's why I had the conversion here, to keep `assignmentSolver` as general as possible. I will still make the change if it's best.

gamemaster/adjudicator/pregame.php
rjmcf: Will be in next commit.

gamemaster/adjudicator/pregame.php
TimothyJones: I like the idea of being robust to future changes - and I think that's a good argument for keeping the same 1-based schema right up to the solver boundary (converting inside the solver). Any changes to what a `countryId` means would be complicated by multiple definitions.

(Also, the code has the assumption of starting from 1, with 0 meaning "everyone" in many many places, so I don't think it'll change)

gamemaster/adjudicator/pregame.php
TimothyJones: Is that commit coming? I'd really like to accept this PR :)

rjmcf:
I've added new commits to fix those two issues, but unfortunately git still sees it as me replacing the whole file. I could do a rebase and submit a new pull request?

TimothyJones:
No need - we care more about what actually changed that what github thinks has changed. I'll accept this, and push it live sometime soon.

rjmcf:
Thank you!

jmo1121109:
This change is awesome. Question for when the announcement is made on it (since I want to explain this well to the community), will taking over a CD position add a record to the table that considers the weighting? And in reverse, what will happen when someone leaves a game via CD that another user takes over, will the first person's country position be considered in future games? If these aren't easily determined questions then don't worry about it, I'll figure it out with some testing once it's live. 

Thank you for all the work you put into this change. Would you have any concerns about being credited on the site's credits page for the development contribution?

TimothyJones:
I don't believe CD takeovers affect this (and I think they shouldn't - you shouldn't be able to "stack your country deck" before a tournament by eg taking over a bunch of Russian positions to lower the chance you draw Russia)

TimothyJones:
I have pushed the change to the server - go ahead and make the announcement, Jmo.

rjmcf:
TimothyJones is correct, however it's possible that the CD code will attempt to do something with the old version of the code which might mess things up. Let me look into it briefly.

Thank you for the kind words! I'd love to be credited, I would have been tempted to ask for it had I known it would be possible!

rjmcf:
From what I can tell CDs won't have any effect on this at all, as desired.

TimothyJones:
I think this is a major contribution, and being named on the credits page is very much appropriate.

Jmo tells me that initial reports from games on the server means this looks like a successful patch!

Thanks again for all your hard work!

TimothyJones:
Let us know what you'd like your credit to read/link to, @rjmcf

rjmcf:
Excellent, glad to hear it's not breaking the site. I'm hoping to make more contributions, so a credit of "Coding Updates and Site Improvements" would be much appreciated. A link to my github profile probably works best for me, my webdip profile leaves a lot to be desired. XD

TimothyJones:
Cool. The new forum has made it pretty easy to turn suggestions into github issues (since they're all in one place). Feel free to grab anything on the issues list, or to create new ones if you want to contribute other changes.

jmo1121109:
Oh you know what, I need to check the country reallocation mod tool and see if that tries calling any of the old code. 

And thank you again rjmcf for taking the time to do all this. I'll be making a PR to the credits page sometime in the next week with a couple of recent contributors and I'll get your GIT linked into that when I do!

rjmcf:
Good spot!

TimothyJones:
@rjmcf: can you comment on #326?

PR #292, opened 2018-02-18
Add Press Check for Mods

jmo1121109:
Adding new tool based on vDip's tool to allow mods to more smoothly
check press with a record in the mod logs.

### DIFF
--- a/admin/adminChatAnalyser.php
+++ b/admin/adminChatAnalyser.php
@@ -1,0 +1,201 @@
+<?php
+
+$countryID1 = $countryID2 = array();
+$gameID=$userID=0;
+$reason = 'Press was checked for gameID: ';
+
+if (!defined('IN_CODE'))
+{
+	chdir ('..');
+	require_once('header.php');
+}
+
+if (!$User->type['Moderator'])
+	die ('Only admins or mods can run this script');
+
+$asTxt = (isset($_REQUEST['TXT'])?true:false);
+$asCsv = (isset($_REQUEST['CSV'])?true:false);
+
+if ( isset($_REQUEST['gameID']) )
+	$gameID=(int)$_REQUEST['gameID'];
+	
+if ( isset($_REQUEST['userID']) && $gameID == 0 )
+	$userID=(int)$_REQUEST['userID'];
+	
+if ( isset($_REQUEST['countryID1']) )
+	foreach ($_REQUEST['countryID1'] as $user)
+		$countryID1[]=(int)$user;
+
+if ( isset($_REQUEST['countryID2']) )
+	foreach ($_REQUEST['countryID2'] as $user)
+		$countryID2[]=(int)$user;
+
+if ( isset($_REQUEST['reason']) && $gameID == 0 )
+	$reason = htmlentities( $_REQUEST['reason'], ENT_NOQUOTES, 'UTF-8');
+
+// Output as Txt or Csv:
+if ((($gameID != 0) && (count($countryID1) > 0) && (count($countryID2) > 0)) && ($asTxt OR $asCsv))
+{
+	$Variant=libVariant::loadFromGameID($gameID);
+	
+	$sql = 'SELECT message, toCountryID, fromCountryID, turn, timeSent
+				FROM wD_GameMessages WHERE
+					gameID = '.$gameID.' AND
+					(
+						( toCountryID IN ('.implode(', ',$countryID1).') AND fromCountryID IN ('.implode(', ',$countryID2).') )
+							OR
+						( toCountryID IN ('.implode(', ',$countryID2).') AND fromCountryID IN ('.implode(', ',$countryID1).') )
+					)
+				ORDER BY id ASC';
+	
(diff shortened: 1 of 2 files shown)


admin/adminChatAnalyser.php
```diff
@@ -0,0 +1,201 @@
...
+$gameID=$userID=0;
+$reason = 'Press was checked for gameID: ';
+
+if (!defined('IN_CODE'))
+{
+	chdir ('..');
+	require_once('header.php');
+}
+
+if (!$User->type['Moderator'])
```
TimothyJones: I think this should be the first line in the file

admin/adminChatAnalyser.php
```diff
@@ -0,0 +1,201 @@
...
+	die ('Only admins or mods can run this script');
+
+$asTxt = (isset($_REQUEST['TXT'])?true:false);
```
TimothyJones: For safety, these should be _POST, not _REQUEST

admin/adminChatAnalyser.php
```diff
@@ -0,0 +1,201 @@
...
+				FROM wD_GameMessages WHERE
+					gameID = '.$gameID.' AND
+					(
+						( toCountryID IN ('.implode(', ',$countryID1).') AND fromCountryID IN ('.implode(', ',$countryID2).') )
+							OR
+						( toCountryID IN ('.implode(', ',$countryID2).') AND fromCountryID IN ('.implode(', ',$countryID1).') )
+					)
+				ORDER BY id ASC';
+	
+	$tabl = $DB->sql_tabl($sql);
```
TimothyJones: This might print errors, and so should be after the header calls

admin/adminChatAnalyser.php
```diff
@@ -0,0 +1,201 @@
...
+					'"'.($from==0?'Global':$Variant->countries[$from-1]).'";'.
+					'"'.addslashes($message).'"'."\n";		
+		else
+			print gmstrftime("%d %b %y %I:%M %p", $timeSent).
+					' ('.$Variant->turnAsDate($turn).')'.
+					' / To '.($to==0?'Global':$Variant->countries[$to-1]).
+					' from '.($from==0?'Global':$Variant->countries[$from-1]).': '.
+					$message."\r\n";		
+	}
+	exit(0);
```
TimothyJones: I think this whole function could be pulled out to a separate file, and called in an If. The exit call would then not be necessary.

admin/adminChatAnalyser.php
```diff
@@ -0,0 +1,201 @@
...
+			print '<input type="hidden" name="countryID2[]" value="'.$user.'">';
+			
+		print  '<input type="hidden" name="gameID" value="'.$gameID.'">
+				<input type="submit" name="CSV" value="Download as .csv (Excel)">
+				<input type="submit" name="TXT" value="Download as .txt">
+				</FORM>';
+	}
+	else
+		print '<br>No sent messages to display.';
+}
```
TimothyJones: Be good to have braces around this else case

admin/adminChatAnalyser.php
```diff
@@ -0,0 +1,201 @@
...
+						( toCountryID IN ('.implode(', ',$countryID2).') AND fromCountryID IN ('.implode(', ',$countryID1).') )
+					)
+				ORDER BY id ASC';
+	
+	$tabl = $DB->sql_tabl($sql);
+	
+	print '<BR><DIV class="variant'.$Variant->name.'">';
+	print '<TABLE class="chatbox">';
+	
+	$alternate = $output = false;
```
TimothyJones: This seems misleading; I think the variables should be declared separately, because they are unrelated.

admin/adminChatAnalyser.php
```diff
@@ -0,0 +1,201 @@
...
+	}
+		
+	while ( list($message, $to, $from, $turn, $timeSent) = $DB->tabl_row($tabl) )
+	{
+		if ($asCsv)
+			print '"'.gmstrftime("%d %b %y %I:%M %p", $timeSent).'";'.
+					'"'.$Variant->turnAsDate($turn).'";'.
+					'"'.($to==0?'Global':$Variant->countries[$to-1]).'";'.
+					'"'.($from==0?'Global':$Variant->countries[$from-1]).'";'.
+					'"'.addslashes($message).'"'."\n";
```
TimothyJones: I’m not sure this is CSV. And, I think that we should escape each field (and addslashes is probably not the right escape function to call). Checkout fputcsv instead.

admin/adminChatAnalyser.php
jmo1121109: Doing so breaks the exports, but I have confirmed that you cannot access any part of this page when not logged in or when logged in as not a mod.

TimothyJones: It is currently the case that you can't use this page while not a moderator, but to avoid bugs in the future, this guard is better placed at the top of the file.

I'm not sure what you mean by breaks the exports?

Issue #295, opened 2018-02-19
Make it clear that game passwords are different from user passwords


TimothyJones:
Currently, a game join page for a passworded game asks for a "password", which is confusing for new users (since it gets confused with the user's password).

We could address this with a simple renaming - perhaps to "invite code", rather than "password".

TimothyJones:
Fixed by #298

PR #298, opened 2018-02-19
Changing Game Passwords to Invite Codes

jmo1121109:
Removing text referring to a game password and replacing it as an invite
code so people no longer think they need to enter their account password
to join a game. Also noticed the country progress color bar was missing
a css 2% margin making it stick out so fixed that too.


jmo1121109:
This addresses https://github.com/kestasjk/webDiplomacy/issues/295

PR #316, opened 2018-05-15
Remove Politics from Preview

jmo1121109:
Removes politics from forum preview, and lets mods or members hide a
thread from the preview using HIDDEN in the thread title.


jmo1121109:
This is currently live on an urgent owner request due to some undesirable conversations looking bad for the site.

PR #317, opened 2018-05-19
Allow the TD of a game to see who's who even in anon games

jmo1121109:
The admins and mods have a long standing rule that you cannot TD a game
you are playing in, so we're allowing TD's to break anon so they can
actually use tools like country swap already available to them.

### DIFF
--- a/gamepanel/member.php
+++ b/gamepanel/member.php
@@ -172,11 +172,12 @@ class panelMember extends Member
 	private $isNameHidden;
 	function isNameHidden()
 	{
-		global $User;
+		global $User, $DB;
+		list($directorUserID) = $DB->sql_row("SELECT directorUserID FROM wD_Games WHERE id = ".$this->Game->id);
 
 		if ( !isset($this->isNameHidden) )
 		{
-			if ( $this->Game->isMemberInfoHidden() && $User->id!=$this->userID )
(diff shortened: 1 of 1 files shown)


gamepanel/member.php
```diff
@@ -172,11 +172,12 @@ function memberCountryName()
...
-		global $User;
+		global $User, $DB;
+		list($directorUserID) = $DB->sql_row("SELECT directorUserID FROM wD_Games WHERE id = ".$this->Game->id);
 
 		if ( !isset($this->isNameHidden) )
 		{
-			if ( $this->Game->isMemberInfoHidden() && $User->id!=$this->userID )
+			if ( ($this->Game->isMemberInfoHidden() && $User->id!=$this->userID) && !(isset($directorUserID) && $directorUserID == $User->id))
 				$this->isNameHidden = true;
 			else
```
TimothyJones: Would this not be better added to isMemberInfoHidden?

jmo1121109: That is what I initially did, but isMemberInfoHidden being turned on for some reason or another allowed a non moderator TD to access the enter account functionality as well as various other moderator tools a TD should not have access too. So I narrowed the focus of it to the function check that only applies to anonymity and made a note to myself to one of these days try and find some time to look more into how isMemberInfoHidden works.

TimothyJones: Oh. That sounds like the function is being misused, and we should probably fix it (or make an issue). I’ll leave the choice up to you, but I’m happy with either.

PR #320, opened 2018-05-19
Fix anon bug

jmo1121109:
Fixing an issue where defeated countries counted as missing orders.

### DIFF
--- a/gamepanel/member.php
+++ b/gamepanel/member.php
@@ -54,7 +54,7 @@ class panelMember extends Member
 		{
 			global $DB;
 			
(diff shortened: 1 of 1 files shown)


jmo1121109:
Due to the wide impact, and after significant local testing, this was pushed live. Apologies for the oversight on this in the initial anon PR.

jmo1121109:
That's a good point. Lets see, I've never adjusted a PR before, but I think I got that to work.

TimothyJones:
Nice work! Thanks!

Issue #196, opened 2015-12-24
Prevent wins during the retreat phase


TimothyJones:
Currently, the territories change hands at the end of the Diplomacy phase, regardless of whether or not there are retreats. This means that it's possible for a country to win at 18 centres when a retreat could have put them back below 17. This should be fixed as a matter of priority.

Ideally the fix would only allow the retreats phase if it is possible for the retreat to put the country back below 18. However, it would still be an improvement to wait until the end of the retreats to change ownership of territories.

This is a change that requires careful testing. Any mistakes when working on the adjudicator could have major consequences to ongoing games.

plund:
See also #261 reporting a very similar issue.

PR #322, opened 2018-05-19
Check for retreat options before win

jmo1121109:
Previously if a player got 18 centers in the fall the game would declare
them the winner before the retreat phase. Now if there is a retreat
phase, the game will wait and check for the winner after retreats are
done.

### DIFF
--- a/gamemaster/members.php
+++ b/gamemaster/members.php
@@ -193,6 +193,8 @@ class processMembers extends Members
 	 */
 	function checkForWinner()
 	{
+		global $DB;
+		
 		/*
 		 * See if only one person is left over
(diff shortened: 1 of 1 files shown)


jmo1121109:
This has been live on vDip for some time now without causing any problems. Local testing was also done, but I will not push this live until I have a game live on the server in the fall season with 17 centers and a retreat into an open center available for testing. If the live test does not work for any reason I'll be rolling back this change immediately due to the wide scope of games it could impact.

jmo1121109:
Will correct https://github.com/kestasjk/webDiplomacy/issues/196

PR #323, opened 2018-05-20
Various Anon Updates

jmo1121109:
After the additional anon protection various members complained it was
now difficult to easily see who had retreat or build options on large
maps. This change will replace the !!, !, grey check, and green check
options with the old forum lock icon to indicate who has moves both in
game and on the member home preview. Only game directors will be able to
see through the lock on the home preview (since TD's spectating games is
quite normal) whereas mods spectating games should not need to see
through (I'm planning a future enhancement to hide members in anon games
for mods who are spectating the game).

### DIFF
--- a/gamepanel/member.php
+++ b/gamepanel/member.php
@@ -471,6 +471,13 @@ class panelMember extends Member
 
 		return '<span class="member'.$this->id.'StatusIcon">'.$this->orderStatus->icon().'</span>';
 	}
+	
+	function memberFinalizedAnon()
+	{
+		if( $this->status!='Playing' ) return '';
+
+		return '<span class="member'.$this->id.'StatusIcon">'.$this->orderStatus->iconAnon().'</span>';
+	}
 
 	private function muteMember() {
 		global $User;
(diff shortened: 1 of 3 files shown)


jmo1121109:
If this one checks out I'll get it up and live over weekend to fix the bugs on the server.

objects/basic/set.php
```diff
@@ -76,6 +76,21 @@ function iconText() {
 		else
 			return l_t('No orders submitted!');
 	}
+	/*
+		Leaving all possible phases in place for clarity on the possible options to make future changes easier.
```
TimothyJones: In general, I'm not a fan of extra code because we think we'll need it later - I think it's better to keep the code cleaner - and that short, clean code is easier to change.

As an example, spliting them out because of future possible changes introduces several places to change if we want to change the lock icon. I think it's better to only split when we need to.

gamepanel/member.php
```diff
@@ -54,12 +54,12 @@ function memberHeaderBar()
 		{
 			global $DB;
 			
-			$row = $DB->sql_hash("select count(1) from wD_Members where gameID = ".$this->gameID." and status not like '%Defeated%' and (orderStatus not like '%Saved%' and orderStatus not like '%Completed%' and orderStatus not like '%Ready%')");
+			$row = $DB->sql_hash("select count(1) from wD_Members where gameID = ".$this->gameID." and status not like '%Defeated%' and status not like '%Left%' and (orderStatus not like '%Saved%' and orderStatus not like '%Completed%' and orderStatus not like '%Ready%' and orderStatus not like '%None%')");
```
TimothyJones: We really need a proper data layer for accessing game info.

TimothyJones:
Some comments inline, but I'll merge this anyway.

objects/basic/set.php
jmo1121109: Okay, so you mean removing the case and just always having this function returning the anon information?

PR #327, opened 2018-05-28
Add isset to isNameHidden

jmo1121109:
The memberBar is called from an external file, which causes various
issues with private variables since they aren't set in that file's
instance. For now, this fix accounts for all cases and renders checks vs
locks for anon for both mods and non mods.

### DIFF
--- a/gamepanel/member.php
+++ b/gamepanel/member.php
@@ -516,7 +516,7 @@ class panelMember extends Member
 	function memberBar()
 	{
 		global $User;
-		if ($this->Game->anon == 'No' || !$this->isNameHidden) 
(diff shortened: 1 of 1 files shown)


gamepanel/member.php
```diff
@@ -516,7 +516,7 @@ private function muteIcon() {
 	function memberBar()
 	{
 		global $User;
-		if ($this->Game->anon == 'No' || !$this->isNameHidden) 
+		if ($this->Game->anon == 'No' || (!$this->isNameHidden) && isset($this->isNameHidden))
```
TimothyJones: There has to be a better way to do this - checking to see if it's set seems like we've made a poor architectural choice elsewhere.

either way, the `isset` check should be before `isNameHidden`.

I'll see if I can take a look for a cleaner solution.

Issue #324, opened 2018-05-28
Colorblind features do not work on mobile display.


mkeCharlie:
# Bug report

Enabling colorblindness features cause the in-game small map to not scale appropriately when viewing on a mobile device.

## Expected behaviour

Entire map is displayed.

## Actual behaviour

Top-left corner of map is displayed.

## Steps to reproduce

1. Enable colorblindness feature (any of the current 3).
2. Navigate to a game on a mobile device.

(Note: Reported on Android v 7.0 with Google Chrome v 66.0.3359.158)

TimothyJones:
Thanks for the report, @mkeCharlie !

cfhughes:
Pull request #328 is a fix for this.

TimothyJones:
Fixed! Thanks!!

PR #328, opened 2018-06-07
Fix colorblind scale

cfhughes:
Fix for #324 
Allows map to scale properly on all screens with colorblindness feature enabled

### DIFF
--- a/javascript/Color.Vision.Daltonize.js
+++ b/javascript/Color.Vision.Daltonize.js
@@ -38,13 +38,18 @@ Color.Vision.Daltonize = function(image, options) {
 		amount = typeof options.amount == "number" ? options.amount : 1.0,
 		canvas = document.createElement("canvas"),
 		ctx = canvas.getContext("2d");
-	if (image.naturalWidth != undefined){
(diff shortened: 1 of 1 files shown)


javascript/Color.Vision.Daltonize.js
```diff
@@ -38,8 +38,8 @@ Color.Vision.Daltonize = function(image, options) {
 		amount = typeof options.amount == "number" ? options.amount : 1.0,
 		canvas = document.createElement("canvas"),
 		ctx = canvas.getContext("2d");
-	canvas.width = image.width;
-	canvas.height = image.height;
+	canvas.width = image.naturalWidth;
```
TimothyJones: I think these properties aren't always available, so we should probably check to see if it's 0 / undefined before using it?

cfhughes: I've added that change. For future reference, is there a list of browsers documented anywhere that webDiplomacy aims to support?

TimothyJones:
Thanks! I slightly changed the guard for safety. Doing the leg work to find the problem is much appreciated!

PR #329, opened 2018-06-09
Submit in game messages without page reload

cfhughes:
Send chat messages in the background, without having to reload page

### DIFF
--- a/javascript/message.js
+++ b/javascript/message.js
@@ -1,5 +1,5 @@
 document.observe("dom:loaded", function() {
-    Event.observe('chatForm', 'submit', function(event) {
+    Event.observe('message-send', 'click', function(event) {
         $('chatForm').request({
             onFailure: function() { },
             onSuccess: function(t) {
--- a/javascript/message.js
+++ b/javascript/message.js
@@ -1,0 +1,15 @@
+document.observe("dom:loaded", function() {
+    Event.observe('chatForm', 'submit', function(event) {
+        $('chatForm').request({
+            onFailure: function() { },
+            onSuccess: function(t) {
+                var cbs = $("chatboxscroll");
+                cbs.update(t.responseText);
+                updateTimestamps();
+                $('sendbox').value = '';
+                cbs.scrollTop = cbs.scrollHeight;
+            }
+        });
+        Event.stop(event); // stop the form from submitting
+    });
+});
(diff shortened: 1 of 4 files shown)


javascript/message.js
```diff
@@ -0,0 +1,15 @@
+document.observe("dom:loaded", function() {
+    Event.observe('message-send', 'click', function(event) {
+        $('chatForm').request({
+            onFailure: function() { },
```
TimothyJones: We should probably do something on failure, like show a message in the chatbox

TimothyJones:
Also, I wonder if it's worth moving to a nicer format for messages between the client and server? (eg json). It might help if we ever moved to websockets.

(Relatedly, @kestasjk you might want to take a look at this PR - I know you had some thoughts on the messaging system)

javascript/message.js
```diff
@@ -0,0 +1,15 @@
+document.observe("dom:loaded", function() {
+    Event.observe('message-send', 'click', function(event) {
```
TimothyJones: Would this be better as `'submit'` rather than `'click'`? That way keyboard shortcuts for submitting forms would still work.

cfhughes: Maybe? I didn't want to have to deal with preventing the default form behavior in that case.

cfhughes: I will try to do some more experimenting with that

cfhughes: I'm having trouble making it work as well with 'submit'. Some older browsers try to keep the default behavior, and end up submitting the form twice.

cfhughes: Are there keyboard shortcuts that currently work?

TimothyJones: https://stackoverflow.com/questions/15612987/stopping-form-submit-with-prototype-js

https://stackoverflow.com/questions/8291805/stop-event-in-ie-9-without-upgrading-to-prototype-1-7

Looks like this is an interaction between earlier versions of Prototype and older browsers. Upgrading prototype could fix it (I don't know if it's backwards compatible, though) - or we could change the way the form is invoked generally.

[Alt] + [Shift] + [S] will submit forms in most browsers on windows, I believe. I have a mac, so can't test it (I don't believe there's an equivalent shortcut by default on mac).

cfhughes: In my humble opinion we should leave it listening to the click event for this change. Perhaps upgrading prototype is a good idea anyway, but I leave that up to you and others who know more than me about the code base as a whole.

[Alt] + [Shift] + [S] didn't seem to work for me.

TimothyJones: I opened #331 to remind us to upgrade prototype, and I'll accept this change. When we apply it to the server we'll have to be careful in case it breaks message sending for anyone.

Thanks very much for this!!

PR #337, opened 2018-10-02
Better loading of message.js to avoid javascript errors

cfhughes:


TimothyJones:
Thanks for this! I'm not sure I follow how it helps - aren't the footer includes still generated at the same time, no matter when we populate the array? Can you explain the improvement?

cfhughes:
This changes it so that the messages.js code is only loaded on the pages it is needed on, rather than all pages on the site. It's a small efficiency improvement, and it also prevents a javascript error that was showing up in the background when that code was loaded on pages other than the pages it's needed on.

TimothyJones:
Thanks for the PR!

PR #341, opened 2018-11-11
Add mod tool temp ban

jmo1121109:
This will add a vDip moderator tool that allows temporary bans on accounts, which will eventually be used to automatically issue temporary bans during CD's.


jmo1121109:
Testing Notes:

Applied changes on local install and issued temporary ban to an account. From the temporary banned account:
Ensured that the notice of being temporarily banned displayed on all pages. 
Ensured that making a new game failed with a reason displayed. 
Ensured that join button was not available for all games in the "new" and "join" sections of the site
Ensured that going directly to a join-able game via url did not display the join button. 
Ensured that re-issuing a temp ban with a time of 0 re-allowed access to join button and making games.

Issue #336, opened 2018-10-02
Misspellings on the American Empire IV map


bmihai:
On both the small and large "Fall of the American Empire IV" maps, the region Coahuila is spelled "Coahulia". The spelling is correct in dropdown menus, but not on the maps.

TimothyJones:
Thanks for the report!

bmihai:
No problem!
Another misspelling, again on both the small and large images of the same map is "Gulf of Fonesca", which should say "Gulf of Fonseca".

TimothyJones:
Thanks for the fix @Espequair!

PR #347, opened 2018-11-29
Fix typos on variant maps

Espequair:


Espequair:
This fixes both typos found in #336

Espequair:
I'll submit only one fix per pull request

PR #348, opened 2018-11-29
fix typos Empire4 (#336)

Espequair:


Espequair:
This fixes the issue raised in issue #336

PR #359, opened 2019-01-06
Add link to new forum

jmo1121109:
Add a link on all the old forum profiles to the new forum profiles of a
user. If the user does not have a new forum profile nothing is shown. If
the server database does not have the new phpbb tables, nothing is
shown.


jmo1121109:
And in horrible coding practice, this is live already. If changes for quality are needed I will make them asap. I choose to test this late at night on a weekend instead of risking it being pushed at a bad time and causing problems.

PR #362, opened 2019-02-27
AdminCP Cleanup

jmo1121109:
Updating the global css to have custom css tags for the admincp and restyling several of the admin cp tabs. Fixing minor issues in the display of status lists as well.

### DIFF
--- a/admin/adminLog.php
+++ b/admin/adminLog.php
@@ -28,9 +28,11 @@ defined('IN_CODE') or die('This script can not be run by itself.');
 
 
 if( !isset($_REQUEST['full']) )
-	print '<a href="admincp.php?tab=Control Panel Logs&full=on">'.l_t('View all logs').'</a>';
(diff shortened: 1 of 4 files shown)


jmo1121109:
To explain in more detail what changed on the status lists page. This page displays all the error logs and all the banned users and donators. On webdip this is an extremely long list, especially when errors have been piling up. So I modified this page to only show the first 50 error logs, and the first 50 banned and donator users and added a link to the top of the page to reload the page without limits using the same style as the status lists page. 

Previously all mod tabs were sharing styling with game tabs, this has been updated on the status lists, admincp header bar, and logs page entirely, and the remaining pages will get new styling soon. This styling was tested out on mobile and desktop. 

Mod Notes was hidden by default on webdip and checked in since the functionality is not fully enabled. Anyone wishing to see it needs to uncomment the line related to it in the admincp.php file in the site root.

Issue #344, opened 2018-11-29
Cancel Games Mod Tool Error


jmo1121109:
Trying to cancel a finished game now causes the following error for admins. I suspect this is due to a change made to the live server during Kestas's fix of the phpmailer. But this will require investigation. 

This error is displayed inline as a result message from the moderator action and does not crash the entire page. 

"Error: The total points transactions (in a finished game) add up to 110, but there are 29 members; cannot cancel game with an unusual points transaction log.

Refunds points each player has bet, and deletes the game.

Game ID: 
228338"

jmo1121109:
Yeah we do need to be able to cancel finished games. Our policy on the team has always been if a user reports cheating in a game before the game is completed and the team finds the game should have been cancelled due to unfair play it is retroactively cancelled to avoid unfairly impacting a member's record simply because the team was slow in getting to the case.

Issue #57, opened 2014-12-10
Mods cannot remove someone from a Pre Game


jmo1121109:
Mods need to be able to remove players from games that have not started yet for a variety of reasons.

jmo1121109:
Done.

PR #364, opened 2019-03-01
Update AdminCP

jmo1121109:
1. update control panel styling.
2. update control panel descriptions
3. changed cancel game to no longer try and refund points on finished games which will remove the issue of cancelled finished games throwing an error.
4. changed force user into CD to allow it to work on a pre-start game by just kicking that person from the game, and if they were the only person in it, deleting the game and resetting the min bet on the game if the weren't. Preventing the issue of mods being helpless on crashed pre-start games.
5. updated the multi finder "active  games" checkbox to say "public games" so it's no longer only looking for active it'll look for people with any public game connections
6. Fix issue in press checker that prevented mods from pulling press in their own finished games.

### DIFF
--- a/admin/adminActions.php
+++ b/admin/adminActions.php
@@ -33,17 +33,18 @@ class adminActions extends adminActionsForms
 	public static $actions = array(
 			'drawGame' => array(
 				'name' => 'Draw game',
-				'description' => 'Splits points among all the surviving players in a game according to 	its scoring system, and ends the game.',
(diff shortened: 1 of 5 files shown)


jmo1121109:
Fixes https://github.com/kestasjk/webDiplomacy/issues/344

jmo1121109:
Fixed https://github.com/kestasjk/webDiplomacy/issues/57

PR #370, opened 2019-03-06
Fix Vanishing Join Button

jmo1121109:
There were several scenarios where a live game could have a player leave or be removed without the min bet on the game being reset, which would cause the game to not have a join button even if it wasn't full.

1. a Mod bans a user in a full pre-start live game.
2. a player in a full pre-start live game leaves.

This has been causing multiple live games to be non-startable despite enough interest.

Adjusted by adding a call to the resetMinimumBet function. This isn't an ideal fix cause it won't be apparent for people working on game code but should work for now.


jmo1121109:
This did not work for all scenarios, investigating further.

Issue #110, opened 2015-02-21
Disappearing pre-game join button


2ndWhiteLine:
Join button disappears occasionally before a game is full.  I believe this happens when the possible seventh player joins at the same time another player leaves.

TimothyJones:
If the last player leaves a game before it has started (eg a live game), then the minimum bet is not correctly reset.

TimothyJones:
This can be fixed with a call to `resetMinimumBet()` when a player leaves a starting game.

jmo1121109:
Fixed

PR #372, opened 2019-03-08
Fix 10 year Issue

jmo1121109:
The reset min bet function was never right for being called as a user was being removed from a pre-start game. The way the logic worked meant the user count wouldn't refect the real user count in the members table. Only discovered when the resetminbet function began getting called when it was initially supposed to. Count is now being pulled direct from the db to avoid state issues.


jmo1121109:
Fixed https://github.com/kestasjk/webDiplomacy/issues/110

Issue #45, opened 2014-11-17
Add ability to find games by name


TimothyJones:
Currently, the only way to find games is with the game search page, which only supports filtering on game features. It doesn't support find-by-name.

We should add the ability to find games by exact match on name.

PR #373, opened 2019-03-09
Add advanced search

jmo1121109:
Adding a new advanced search tool to the site that allows for complex searches on games and users. Currently only going to be pushed live for mods to test load balances and a couple features that will only work live.


jmo1121109:
Solved https://github.com/kestasjk/webDiplomacy/issues/45

Issue #70, opened 2015-01-04
Search Paramater Update


jmo1121109:
"Hours per phase" should be renamed "Time per phase". Since not all games are at least 1 hour.

jmo1121109:
Fixed

Issue #54, opened 2014-12-08
The joinable games list currently shows several games which are not actually joinable


TimothyJones:
For some reason, the joinable games tab shows games which are not actually joinable (ie, they have no open positions). Perhaps it is showing games which used to have joinable positions, or were joined this turn?

jmo1121109:
Fixed.

PR #382, opened 2019-03-16
Fix game search issues

jmo1121109:
Rename hover over text on "Joinable" to inform that it's public games only.

Redo wD_Misc GamesOpen query to correctly pull the number of joinable games. Old query was incorrect.

Rename "Hours per phase" in search criteria to "Time per phase" since all phases aren't hours like live games and 10 day games.


jmo1121109:
Fixes https://github.com/kestasjk/webDiplomacy/issues/70

jmo1121109:
Fixes https://github.com/kestasjk/webDiplomacy/issues/54

PR #383, opened 2019-03-16
Live games show timestamp on new game selection >1 day until start

shdant113:


jmo1121109:
To provide some more information on this since it was a joint effort. Currently games that are days in the future from starting or having a turn process do not display the hour/minute that they will process or start. So for live games, which require a player being online at the start time, it's near impossible to start them days in advance. Finished games and WFO games also do not show the time of completion or the expected process time in hours/minutes so this will add that information to those locations as well.

jmo1121109:
The php time format was changed in addition to the javascript format for the rare cases where users javascript is not enabled so they'll still see a UTC date with the information they need to do a time zone conversion.

Issue #116, opened 2015-03-05
Make it possible to look at games that you have played in with another player 


Sleepcap:
1) Make it possible to look at games that you have played in with another player (e.g. if you go onto their player profile, just a button which then shows any games which you both played in). I only propose this on the basis that I imagine (though I have no programming skills) that it would be relatively easy to implement.

jmo1121109:
Done.

PR #385, opened 2019-03-17
Enhance advanced search

jmo1121109:
Allow people to search for a given user's games and check for games between a user and themselves. Finished games only.

Update the "show only joinable games" in the game search to filter out "Pre-start" games. 

Refactor some of the css tags and code, add more comments.


jmo1121109:
https://github.com/kestasjk/webDiplomacy/issues/116

Issue #391, opened 2019-03-19
Add Hover Information to Profile "Rank"


jmo1121109:
The profile "Rank", ex: Rank: Diplomat, does not explain how this information is determined. Add a hover over to the rank explaining what the status signifies.

Top 1% of points, over 1,000 points, top 5% win%. Whatever the criteria for the rank actually is, this is a commonly asked question by members.

jmo1121109:
Done.

Issue #390, opened 2019-03-19
Fix Profile Mobile View


jmo1121109:
On an Iphone SE screen with a username of 10 or more characters the username is not wrapped, so the entire left hand column of the profile is shoved below the right column. Wrap this element to prevent the left column from being shifted down.

jmo1121109:
Done.

PR #396, opened 2019-03-21
Profile Page Improvements

jmo1121109:
1. Fix username to wrap on left column of profile.

2. fix moderator disclaimer link to wrap

3. add Classic and Variant stats to the profile page.

4. Add hover over that explains what the different "ranks" mean...somewhat, at least more then now.


jmo1121109:
Fixed https://github.com/kestasjk/webDiplomacy/issues/391
Fixed https://github.com/kestasjk/webDiplomacy/issues/390

PR #399, opened 2019-03-22
Removed columns and changed format to declutter the credits page

shdant113:
Removed the columns from the credits page and rearranged the information displayed so that it can be easily accessed and updated with relevant information as necessary without being more cluttered. Added basic styling for the new credits page as well but without a local environment to test on I cannot see it, so the style changes may not be quite right yet.


jmo1121109:
Thank you for doing this!

PR #429, opened 2019-04-12
Create a user option to sort possible orders in dropdown

Squigs44:
Create a user option to sort the dropdown options when loading possible orders. Defaults to sorting alphabetically, with options to revert to no sort and to push all the via convoy orders to the bottom (while sorting alphabetically still).


jmo1121109:
Very nicely done, approved.

Issue #301, opened 2018-02-25
Max Game Length Name With Repeat Name Causes Error


jmo1121109:
The site has code in place to adjust game names for new games if a game of the same name already exits. 

For example, creating a game named "test" results in a game named "test-64" since 63 versions of test have already been made. 

The site also has code to truncate game names that are too long. If you enter "test test test test test test test test test test test test test test test test test test test test test" the game name is truncated to fit in the database column"

But if you enter a very long name that has already been used like "buffalo buffalo buffalo buffalo buffalo buffalo buffalo buffalo buffalo buffalo buffalo buffalo buffalo buffalo buffalo" the site truncates the name  to fit the database column and THEN appends the "-2" which makes the name too long for the site and you get the following error. 

"Error triggered: Duplicate entry 'Buffalo Buffalo Buffalo Buffalo Buffalo Buffalo Bu' for key 'gname'

Error: "Duplicate entry 'Buffalo Buffalo Buffalo Buffalo Buffalo Buffalo Bu' for key 'gname'"
Raised: "/usr/local/www/apache24/data/webdiplomacy/objects/database.php"
Line: "377"
userID = 15658Variable dump: sql => INSERT INTO wD_Games
SET variantID=1,
name = 'Buffalo Buffalo Buffalo Buffalo Buffalo Buffalo Buffalo Buffalo Buffalo Buffalo Buffalo Buffalo Buffalo Buffalo Buffalo Buffalo Buffalo Buffalo Buffalo Buffalo Buffalo ',"

TimothyJones:
Good catch. Swapping the order of the numbering and the truncation would fix this in the short term. Consequently, I'm going to mark as `good-first-issue`.

Longer term, it would be good to have a more principled approach (like having a pre-defined game name length that was a bit longer but also reported in the user interface).

jmo1121109:
Fixed

PR #431, opened 2019-04-12
Fixed game name uniqueness/length bug

Squigs44:
Addresses issue 301. When renaming for uniqueness, the new game name will now be guaranteed to stay less than or equal to 50 characters so that it fits in the database.


jmo1121109:
Fixes https://github.com/kestasjk/webDiplomacy/issues/301

Issue #430, opened 2019-04-12
User info page when registering needs several updates


Squigs44:
While registering, the page that takes in user information has several issues. First, the options that it collects during registration do not save. Secondly, all the <hX> tags need to have the indent overwritten on all of the various registration step displays. Finally, the readonly "email" field shown with the registration token needs to be grayed out.

jmo1121109:
Thanks for this report, verified, and planning on addressing these this weekend.

jmo1121109:
Fixed

PR #436, opened 2019-04-13
Various Registration Fixes

jmo1121109:
1. Fix user registration bug that ignored all radio button user options on account creation.
2. Fix multiple styling bugs on user registration
3. Remove page by page styled submit buttons for the new default button
4. Update Register page to add warning to avoid personal information in username and to keep username appropiate.


jmo1121109:
Fixes https://github.com/kestasjk/webDiplomacy/issues/430

Issue #278, opened 2017-12-21
Move the JS and CSS version parameters to global/definitions.php


Sleepcap:
At the moment the parameter of the CSS-version is set directly in the code with the "&ver=" parameter in html.lib and the JS-Version is set mid-code too.
Maybe make this a GLOBAL variable, so it's easier to manage.

Sleepcap:
Merging the code today I realozed the JD-Version is now defined 2 times in html.lib (line 390 and 920).
This should really move to global/definitions.php.

jmo1121109:
This will need to include adding css versioning to all the css files too. Currently we only css cache bust some of the files.

Sleepcap:
You do the CSS-versioning in the lib/html.php file too..
Would be much better defined in definitions.php.
It's really cluttered at the moment and makes updating the vDip-code unneccessary complex.

PR #438, opened 2019-04-14
Version Control

Squigs44:
Creates 2 global variables in global/defintions.php that controls all versioning for all js and css files.

Fixes https://github.com/kestasjk/webDiplomacy/issues/278


jmo1121109:
Looks good, thank you for this!

Issue #137, opened 2015-04-24
Create a database install script that contains the latest version of the database for a fresh install


TimothyJones:
Currently, after the DB install script is run, each `update.sql` under the `install` directory needs to be run in order. It would be a dramatic improvement if the `install.sql` file were (automatically) kept up to date with the database updates.

jmo1121109:
Fixed

PR #448, opened 2019-04-21
Create Full DB Install

jmo1121109:
This is a full install of the base DB with every update that will get a test environement up and running instantly.


jmo1121109:
Addresses https://github.com/kestasjk/webDiplomacy/issues/137

Issue #265, opened 2017-11-04
NMR wrongly credited on ready and unready of holds


negaffney:
# Bug report

When a player readies all holds and then unreadies them (leaving them saved), and does not save any other orders at any point during that phase, the player is credited with an NMR in their reliability statistics. 

## Expected behaviour

The player should not be credited with an NMR in such instances. 

## Actual behaviour

The player is credited with an NMR. 

## Steps to reproduce

In a new phase of any game, click ready immediately without editing any orders, and then click unready and leave the orders alone for the rest of the phase.

TimothyJones:
Nice catch. This might be annoying to fix due to the way NMRs are detected, but we should definitely fix it.

jmo1121109:
Fixed by bo's pr.

PR #450, opened 2019-04-25
Fixing bug that assigns user NMR when they follow auto-input orders

shdant113:


jmo1121109:
Fixes https://github.com/kestasjk/webDiplomacy/issues/265

PR #452, opened 2019-04-30
API Integration

ppaquette:
Still a work in progress...

Things left to do:

- [x] get_game_state() should return a list of phases with 'units', 'centers', 'orders'
- [x] get_game_state() should return a "standoff" with list of locs in standoff to properly compute retreat locs
- [x] set_orders not working for convoy orders
- [x] set_orders not working for move via convoy orders
- [x] get_game_state() does not return units and orders for Builds phase
- [x] All orders are ignored if one of the orders is invalid
- [x] Work around multi-account detection mechanism
- [x] 7 bots all ready doesn't make the game advance automatically
- [x] If a bot submits an order for its own account through the API, it should not be marked as in civil disorder for that turn.
- [x] Units for previous Builds/Retreats phase are incorrect
- [x] Check permission when sending orders for a user in CD
- [x] Make sure that orders for a user in CD don't remove the CD status from the user
- [x] Debug to make sure everything is working properly


jmo1121109:
It will take me a little bit of time to get through reviewing this due to the size but this is fantastic and I'll make sure to get you review comments as soon as I can. 

You can check off, Work around multi-account detection mechanism, since we will just instruct the moderator team to ignore the connections that will appear as a result of your use of several accounts. If you need to make accounts for testing a live game let me know and I can get those flagged for moderators to ignore. 

When you get to the test game step I can also make you the "Game Director" of that game which will allow you to speed up the processing of each turn significantly.

ppaquette:
@jmo1121109 I'm done fixing bugs. This should be ready to review.

api.php
```diff
@@ -0,0 +1,773 @@
...
+		$gameID = $args['gameID'];
+		if ($gameID == null)
+			throw new RequestException('Game ID not provided.');
+		$gameID = intval($gameID);
+		$Variant = libVariant::loadFromGameID($gameID);
+		libVariant::setGlobals($Variant);
+		$gameRow = $DB->sql_hash('SELECT * from wD_Games WHERE id = '.$gameID);
+		if (!$gameRow)
+			throw new RequestException('Invalid game ID');
+		return new Game($gameRow);
```
kestasjk: Shouldn't this return $Variant->Game() or some such? I remember that you need to request all new objects from the variant layer to give the variant a chance to modify the game.

api.php
```diff
@@ -0,0 +1,773 @@
...
+}
+
+try {
+	// Load API object, load API entries, parse API call and print response as a JSON object.
+	$api = new Api();
+	$api->load(new ListGamesWithPlayersInCD());
+	$api->load(new ListGamesWithMissingOrders());
+	$api->load(new GetGamesStates());
+	$api->load(new SetOrders());
+	$jsonEncodedResponse = $api->run();
```
kestasjk: Very neat

api/responses/game_state.php
```diff
@@ -0,0 +1,437 @@
...
+	public $standoffs = array();
+
+	/**
+     * List of occupiedFrom territories
+     * @var array
+     */
+    public $occupiedFrom = array();
+
+	/**
+	 * Load the UserOptions object. It is assumed that username is already escaped.
```
kestasjk: GameState

api/responses/members_in_cd.php
```diff
@@ -0,0 +1,76 @@
...
+	function load()
+	{
+		global $DB;
+
+		$countryTabl = $DB->sql_tabl("SELECT m.gameID, m.countryID
+                                      FROM wD_Members AS m
+                                      LEFT JOIN wD_Orders AS o ON ( o.gameID = m.gameID AND o.countryID = m.countryID )
+                                      LEFT JOIN wD_Games AS g ON ( g.id = m.gameID )
+                                      WHERE NOT o.id IS NULL
+                                            AND g.variantID in (1, 15, 23)
```
kestasjk: Eek, the variant ID can’t really be counted on to be the same from server to server. Can this be added to Config::$something ?

api/responses/members_in_cd.php
```diff
@@ -0,0 +1,76 @@
...
+                                            AND g.phaseMinutes >= 10
+                                            AND m.missedPhases > 0
+                                            AND g.processStatus = 'Not-processing'
+                                            AND g.phase IN ('Diplomacy', 'Retreats', 'Builds')
+                                            AND g.processTime >= UNIX_TIMESTAMP()
```
kestasjk: UNIX_TIMESTAMP() looks unfamiliar, can we use time() or whatever is used elsewhere in the code? Otherwise there just might be some weird time zone business if the database defaults to something different. (Unlikely, but a quick fix)

api/responses/unordered_countries.php
```diff
@@ -0,0 +1,72 @@
...
+	function load($userID)
+	{
+		global $DB;
+
+		$countryTabl = $DB->sql_tabl("SELECT m.gameID, m.countryID
+                                      FROM wD_Members AS m
+                                      LEFT JOIN wD_Games AS g ON ( g.id = m.gameID )
+                                      WHERE (m.orderStatus IS NULL OR m.orderStatus = '')
+                                            AND m.userID = $userID
+                                            AND g.variantID in (1, 15, 23)
```
kestasjk: Can this be configurable please as above

api/test.php
```diff
@@ -0,0 +1,179 @@
...
+	 */
+	private $apiKey;
+
+	/**
+	 * Initialize an ApiCall object.
+	 * @param string $siteUrl - URL of a WebDiplomacy instance.
+	 * @param string $apiKey - API access key/
+	 */
+	public function __construct($siteUrl, $apiKey) {
+		$this->siteUrl = trim($siteUrl, " /\t\n\r\0\x0B");
```
kestasjk: Can you add a comment for what this does? It looks like the kind of thing that might be different from webserver to webserver. Not the end of the world, but it looks .. delicate

kestasjk: (For test code I’m not too bothered, but I saw this above I think)

api/test.php
```diff
@@ -0,0 +1,179 @@
...
+				'orders' => $orders,
+				'ready' => $ready
+			));
+	}
+}
+
+/**
+ * Usage example of ApiCall.
+ * Make call to all current API entries.
+*/
```
kestasjk: This definitely needs to be disabled and only active if a Config:: value is set please

board/orders/diplomacy.php
```diff
@@ -155,6 +155,11 @@ protected function viaConvoyCheck()
 	protected function checkConvoyPath($startCoastTerrID, $endCoastTerrID, $mustContainTerrID=false, $mustNotContainTerrID=false) {
 		global $DB;
 
+        if ($this->convoyPath == [-1]) { // API - We are offloading computing the convoy path to the server
```
kestasjk: Mmmm... this isn’t the code that computes the path this is the code that checks it. This would let anyone submit a convoy path of -1 and their submitted convoy wouldn’t work. This looks like a security issue unless I’m mistaken; the system needs to check this please.

kestasjk: *Their submitted convoy /would/ work, even if it was convoying from Clyde to Turkey with nothing in between, I believe the adjudicator assumes convoy paths are valid.

api/responses/members_in_cd.php
jmo1121109: Oddly enough, because the variants are made by 1 person, and the variants are then ported to other sites with the variantID defined in the variant install files they're actually identical on all 4 of the webdip code base sites as far as I've been able to tell, so I've unfortunately used this logic in a couple spots. I'll take it upon myself to see if I can come up with a better solution but I'm okay with this in here, and I'll work with squigs44 to get a better design for the few spots we've done this. Sorry about that. 

Definitely agree on the the need for a config setting to be able to entirely turn off the api/ensure a site without it setup won't have problems though.

api/responses/unordered_countries.php
jmo1121109: I'm not as clear on what data this query is trying to pull, if you explain it in English I can confirm if the right columns are in play here.

ppaquette: It lists the game a user is currently playing that have no orders.
This way, the bot can join a game, find that it needs to submit orders for the current phase, and then wait until it needs to submit orders again in the next phase.

SELECT m.gameID, m.countryID
--> Returns the gameID and countryID where the orders as missing

FROM wD_Members AS m
LEFT JOIN wD_Games AS g ON ( g.id = m.gameID )
--> Using the order status from the members table

WHERE (m.orderStatus IS NULL OR m.orderStatus = '')
--> where the orders have not been submitted yet

AND m.userID = $userID
--> JOIN member table on game table

AND g.variantID in (1, 15, 23)
--> Only for variants that are supported by the API

AND g.processStatus = 'Not-processing'
AND g.phase IN ('Diplomacy', 'Retreats', 'Builds');");
--> Only for active games

api/responses/members_in_cd.php
```diff
@@ -0,0 +1,76 @@
...
+	 */
+	function load()
+	{
+		global $DB;
+
+		$countryTabl = $DB->sql_tabl("SELECT m.gameID, m.countryID
+                                      FROM wD_Members AS m
+                                      LEFT JOIN wD_Orders AS o ON ( o.gameID = m.gameID AND o.countryID = m.countryID )
+                                      LEFT JOIN wD_Games AS g ON ( g.id = m.gameID )
+                                      WHERE NOT o.id IS NULL
```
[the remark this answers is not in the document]
ppaquette: This SQL query is trying to find any members (either active or CD) that have not yet submitted orders on any game that will be processed in the next 60 secs.

The missedPhase attribute is not touched, therefore that player will trigger the CD logic at the next process() (within 60 secs) and will be marked as left, therefore someone else can take its place.

If I wait until the player is marked as "Left", there is at least one turn where no orders will be submitted at all. Let me know what logic you prefer.

The model should give reasonable moves even for "Press" games, because it uses the previous orders as a proxy for relationships. I'm not sure we need to add the 'NoPress' clause, but I can add it if you want.

I'll add the 'DISTINCT' keyword and remove the 10 mins minimum phaseLimit clause.

api.php
ppaquette: `game.php` seems to set `$this->Variant` using the value of `$GLOBALS['Variant']` which is already set at L250.

```
		$gameID = intval($gameID);
		$Variant = libVariant::loadFromGameID($gameID);
		libVariant::setGlobals($Variant);
```

Do you think that should work, or is there a specific method I need to call?

ppaquette:
Thanks for the reviews.

I've incorporated all the requested changes, added better documentation to the SQL queries, and rebased the commits.

Let me know if there are other things you want me to fix.

api/responses/unordered_countries.php
jmo1121109: Okay gotcha, so even if you take over for a user, you might not be taking over for all of their games. If a user has a 10 day phase and a 1 day phase game. They can be in CD in the 1 day phase game with another 6 days before they'll even be at risk of CD'ing in the 10 day phase one, even if they don't have orders entered. So you'll want to add in AND m.status = 'Left'

api/responses/members_in_cd.php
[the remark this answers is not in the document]
jmo1121109: Oh okay I understand the confusion now. So in the past couple weeks I pushed live a change that added "ExcusedMissedTurns" to games. So each game now has 0-4 missed turns. The default is 1. If a player does not have orders entered when the turn goes to process they are charged a missed turn and the phase resets time resets to prevent the case where no orders will be submitted. If the player does not have an excuses left they are kicked into CD and the phase time resets, meaning that you'll be able to only look for people who have a status of left, and pick moves for them, without causing games to process with missing turns. Your code and this recent change should go really nicely hand in hand to make the game flow smoother for members. 

For the initial launch I would like to limit it to no press games, if there aren't enough games that meet the criteria then we should be able to setup some games with moderator"dummy" accounts taking up 1 position, force them into CD and allow the players to play against 1 of the bots in a private game to ensure nobody takes the position over. And then once we've had some successful games go through we can roll this out to press games. I'll just want a little time to put a measure in place to prevent this from happening on some press games that are apart of tournaments or other special event games. So I'll get something in place for that asap. 

Let me know if you need more details on the recent change adding missedTurns, but I tried to make sure it wouldn't conflict with your code at all.

api/responses/unordered_countries.php
ppaquette: There is probably still some confusion here.

There are two components:

1) Bot temporary submitting orders when a player has not submitted their orders in a phase

- SQL query in file "members_in_cd.php"
- Orders can be submitted by bot if no orders have been submitted by a player and there is less than 60 secs for a player to submit their orders before the next process.

2) Bot is playing a game as a regular user (e.g. 1v1 or tournament)

- SQL query in file "unordered_countries.php"
- In this case, we find all the games where the bot user ID is playing that have no orders submitted yet
- There is no need to add "m.status = Left" since we are playing as a regular user.

jmo1121109: Okay, now I actually gotcha. So in this case for the members_in_cd, my comments about adjusting based on the new excused missed turns making status = 'Left' the correct method are still correct. For the unordered_countries file disregard my comments. Thanks for clarifying the intent between the two.

jmo1121109:
I've been sick this past week but I'll try to get to reviewing the changes this weekend. Sorry about the delay.

ppaquette:
Thanks for merging this.

Next steps should be:

1) Deploy to live site
2) Create a bot account with a random password (e.g. "KestasBot")
3) Create an API key for the bot account
4) Assign permissions getStateOfAllGames, submitOrdersForUserInCD, and listGamesWithPlayersInCD to that API key
5) Send me the API key by email with the public SSH keys for the users you want to have access to the bot server
6) I'll setup the server on my side with the API key
7) Debug by creating a single no-press 6 humans vs 1 bot game, to make sure the bot can play as itself and can replace users in CD
8) Debug the 1v1 maps
9) Gradually add more games

jmo1121109:
Hey Philip, 

So I have this code deployed live after doing a lot of local testing just to make sure everything else still ran as expected. I can make the bot for step 2. Can you walk me through steps 3 and 4? I want to make sure I do those correctly. 

Thanks!

ppaquette:
1) Login to the admin control panel
2) Go to "API - Add an API key for a user" and enter the user id of the bot

Selection_053

3) You want to write down the API key created.

Selection_054

4) Go to "API - Set API key permission" and set "getStateOfAllGames", "submitOrdersForUserInCD", and "listGamesWithPlayersInCD" to "Yes"

Selection_055
Selection_056
Selection_057

5) Go to "API - Show API key and permissions for a user" and enter the bot's user ID. This should show you the API key and the 3 permissions set to "Yes"

Selection_058
Selection_059

Issue #63, opened 2015-01-03
Site banner Redesign


jmo1121109:
Redesign the site banner to operate on a table, with a mod option to update. This removes the rather dangerous design of altering the webconfig file to change the banner.

jmo1121109:
Fixed by @Squigs44 pr. Thanks for that.

PR #462, opened 2019-05-07
Banners

Squigs44:
Decided to make a whole new branch and PR just to make the merging easier. Retested everything with all the new stuff and it all checks out. Also fixed the panic message so that it now takes html tags. Got rid of serveroffline from the install file since it doesn't do anything. Also I think I fixed the bracket styling. And I changed the name of the mod tool to change the notice message so that alphabetically all three are next to each other.


Squigs44:
Fixes https://github.com/kestasjk/webDiplomacy/issues/63

PR #478, opened 2019-05-13
Chat archive

Squigs44:
Fixing the bug with chaos chat archive


Squigs44:
So there was a bug which was breaking the chat archive, which caused the first page to show all of the messages, the second page to show all of the messages except what was supposed to be on the first page, and so on. The chaos variant previously had a class extension on the chatbox class which multiplied the limit by 5. This worked for normal chatbox use, since the limit was an int. However, the archive pager used the chatbox class, and the limit was a string rather than an integer, so multiplying by 5 broke the paging. We fixed this by simply removing the class extension so that the chatbox functions for chaos the same way it does for all the other variants on webdip.

Issue #59, opened 2014-12-10
Cancelling a game does not make a record in the Control Panel Logs


jmo1121109:
The control panel log would ideally keep a record of the user id's of the members in the game at the time it was cancelled, the person who cancelled it, time cancelled, and a reason for cancelling.

jmo1121109:
Update on this, this works in my dev environment but not on the live site. So this will require some in depth investigating.

PR #479, opened 2019-05-14
Make a log in the adminlog when cancelling games

Squigs44:
Currently the way that cancelling a game works is slightly broken which means cancelling games doesn't make a log in the admin log. I hard-coded a work around for this, with more details being added into the log.


Squigs44:
Fixes https://github.com/kestasjk/webDiplomacy/issues/59

Issue #200, opened 2015-12-28
Correct the message archive direction


HellenicRiot:
Currently game messages archive linked at the bottom of the game page displays posts in reverse chronological order, which is annoying. It would be more intuitive if they added new messages on at the end and could be read in chronological order.

PR #506, opened 2019-05-25
Message Archive Paging

Squigs44:
Redoes the paging for archived messages in a game to be uniform with the other paging changes made recently. Also adds an option in the archived messages to filter down to just a single country, or global, or notes.

Also fixes issue https://github.com/kestasjk/webDiplomacy/issues/200 from 2015, after realizing that the messaging ordering gets switched several times in order to display the recent messages in the correct place.


Squigs44:
The messages.php file changed in this PR is only used for the message archiving and basically only deals with the paging. It utilizes the Chatbox class, which I know the World and Chaos variants modify. Can you give me an instance of a specific variant, how it changes what is in this specific file, and how you currently do this?

I know that there is at least one other file named message.php in the code, one that isn't buried in board/info, which I think deals with the main messaging, and it also utilizes the Chatbox class.

Sleepcap:
Ahh sorry about this.
I was talking about the orderarchive....

PR #529, opened 2019-06-02
Missing ";" prevented a clean fullinstall

Sleepcap:
Small error I found during the setup of my new webdip installation.


jmo1121109:
Thanks for that!

Issue #535, opened 2019-06-10
Reliability-page fails to load if there are NMRs in cancelled games


Sleepcap:
If you open the reliability-breakdown on a user and some of the games are cancelled it fails to load the explanation-page.

jmo1121109:
Do you have an example profile on vDip I can check this out on?

Sleepcap:
Check userID=12023 on vDip

jmo1121109:
Fixed.

PR #537, opened 2019-06-13
Fix Bug on RR Display

jmo1121109:
If a user had a missed turn in a game that had been cancelled the RR display page was breaking. Fixing the variant setting and variant related information to be inside a conditional if the game still exists and just showing base info on the cancelled game's missed turns.


jmo1121109:
Fixed https://github.com/kestasjk/webDiplomacy/issues/535

PR #550, opened 2019-06-21
Typos

firexe:
Ignore the credits page, haha - but everything else is just typos and grammar issues.


variants/AncMed/rules.html
```diff
@@ -1 +1 @@
-Unlike the original variant-rules a fleet occupying the Baleares may <b><u>not</u></b> be used to convoy an army using the normal convoy rules.<br><br>
+Unlike rules of the original variant, a fleet occupying the Baleares may <b><u>not</u></b> be used to convoy an army using the normal convoy rules.<br><br>
```
jmo1121109: This sounds slightly awkward, can you update that to "Differing from the rules of the original variant, a..."?

locales/English/intro.php
```diff
@@ -113,7 +113,7 @@
 
 	<li class="formlisttitle">Move</li>
 	<li class="formlistdesc">
-		The unit tries to move into(/attack) an adjacent territory.
+		The unit tries to move into (or attack) an adjacent territory.
```
jmo1121109: into (also know as attacking) an adjacent
might be better since they're the same thing, not different.

locales/English/credits.php
```diff
@@ -57,7 +57,8 @@
 	array(
 		'The webDiplomacy Developers',
 		'<a href="/profile.php?userID=15658">jmo1121109</a>, current lead developer<br />
-		<a href="/profile.php?userID=10">kestasjk</a>'
+		<a href="/profile.php?userID=10">kestasjk</a><br />
```
jmo1121109: Can you just remove this file from your PR? Aurelin mentioned you weren't fixing this, so I already did, sorry!

firexe: Yep yep. 
Although, I do have to ask, why change to green? (and not keep the grey, haha)

locales/English/intro.php
firexe: "(or if it's occupied, attack)"
Seems to match better?

locales/English/credits.php
jmo1121109: bo_sox48 is working on an optional dark mode, and the green looks better on both than the grey does.

firexe:
Not sure, is it back?

firexe:
Happily admitting I've got no clue what I'm doing :D

jmo1121109:
It is back, but git doesn't seem to know that there's no differences between your proposed file and the live one and I have no idea how to fix that unfortunately.

PR #551, opened 2019-06-21
Typos2 (this ought to work)

firexe:
This ought to work now :D
Do tell me if I missed anything (compared to original), but don't think I have.


jmo1121109:
Thanks for this!

PR #553, opened 2019-07-10
Dark Mode

shdant113:
Users can toggle between traditional light mode theme and dark mode styles


Sleepcap:
I'm not sure if it's a good idea to duplicate all the css files with a different color-scheme.
That way everytime someone changes or adds something to an existing css he needs to make sure he does this in all profiles.
Maybe overwrite the colors like I do with the vDip css.

detailedSearch.php
```diff
@@ -685,13 +685,23 @@ class GameResultData
 		if ($seeBronze=='checked') { print '<th class= "advancedSearch">IsBronze</th>'; }
 
 		print "</tr>";
-
+		global $User;
```
jmo1121109: Can you move this to the top of the function just to match how it's used everywhere else?

global/definitions.php
```diff
@@ -48,7 +48,7 @@
 
 //Increment these versions whenever you update any js or css files for cachebusting
 define("JSVERSION",1.2);
-define("CSSVERSION",1.23);
+define("CSSVERSION",1.24);
```
jmo1121109: I think you need to update the version to 1.57 if you're pushing that same SQL version. We're on 1.56 right now.

objects/basic/set.php
```diff
@@ -53,12 +53,15 @@ class setMemberOrderStatus extends set {
 	 * @return unknown_type
 	 */
 	function icon() {
+		global $User;
```
jmo1121109: Is any functionality actually being added in this file?

objects/user.php
```diff
@@ -1272,5 +1298,16 @@ public function userIsTempBanned()
 
 		return $tempBan > time();
 	}
+
+	/* 
+	 * Determine style theme user is using
```
jmo1121109: Mention in the comment that this will be null in some cases so additional handling is needed for those cases, or just adjust this to return light mode if the query didn't return anything and remove the handling from the calling functions.

PR #587, opened 2019-08-04
New forum icons for dark mode

firexe:
I think all the currently used images have had the white edited out, unless I've accidentally missed something. The only issue is, with the actually animated gifs (the four with "_hot"), I've ran into an issue with preserving transparency in the final product - as a result, they're a bit subpar due to having to compensate for both backgrounds. 

If the outcome is still too jarring on a dark background, let me know and I shall try something new.


jmo1121109:
Thanks for this, I'll be able to get these live sometime soon. Due to how phpbb is setup I'll have to manually copy them to a different directory.

Issue #227, opened 2016-02-24
Ordering nonsensical convoys doesn't provide clear feedback to the user


TimothyJones:
If we have a classic game with:

> A Bre
> F MAO
> F NAO

and we try to order NAO convoy Bre->Spain (an unnecessary order), then the convoy order partially disappears as the javascript validator doesn't handle this case gracefully. The orders are then left in a state where they can't be saved.

This is particularly a problem for the World map, since nonsensical convoy orders are possible much more often (and with similarly named territories, can happen easily).

We should alter the javascript to prevent nonsense convoys from being ordered.

yewang:
Maybe the validator should not try to eliminate nonsensical convoys, and instead just let them be ordered. When looking at only one set of orders, it can be impossible to tell if a convoy order is nonsensical/unnecessary or if it is indeed a vital segment of an internationally coordinated convoy.

Further, it is sometimes nice to be able to order nonsensical convoys to communicate to other players in gunboat, and reducing the validator complexity may speed things up and avoid bugs.

What are your thoughts? I can take a look at the javascript to see how to address this.

kestasjk:
Is Bre-Spain really unnecessary? I vaguely remember in the DATC tests there were some cases involving bypassing a head to head because of a convoy possibility.

TimothyJones:
The example nonsensical order here is NAO covoy Bre->Spain, since that fleet can't ever be used meaningfully in the convoy.

@yewang: I suspect the adjudicator relies on some of the pruning that the validator does - @kestasjk, do you know if this is correct?

kestasjk:
Ah I see, my territory acronyms are rusty. ;)

The adjudicator does assume orders are valid, but to clarify (as you probably already know) the server-side code does not assume that the client-side order validation has been run. The client will submit a convoy chain for any submitted convoy order, so that the server can quickly validate the chain rather than search for a valid chain itself.

Looking at the JS now, and bearing in mind I'm very rusty, here is rough code which I think would prune out invalid orders:
[https://github.com/kestasjk/webDiplomacy/blob/master/javascript/board/model.js#L241](url)
`getConvoyFromChoices : function(ToTerritory) { 
 if( this.convoyLink ) 
 { 
	this.convoyOptions = this.ConvoyGroup.Armies.select(function(a){return a.Territory!=ToTerritory;}).pluck('Territory').select(function(fromTerritory) {
	// Ensure there is a valid path from this possibility to ToTerritory via the fleet territory:
[...]

Issue #189, opened 2015-10-23
Busy script on world maps with lots of fleets in the ocean


TimothyJones:
There is an issue with the javascript for calculating convoys that causes unresponsive script (and the inability to enter orders) when the world map has many fleets in the ocean.

This is probably responsible for webdip using a lot of battery on mobile, too.

I'm not sure what the problem is, more investigation is needed.

diptobi1:
Just a notice I might have solved this issue by rewriting and optimizing the algorithm to find Convoy paths: 
The new code has just been included on vdip and might be a good candidate to be ported over here, too, after some time of testing on vdip. Tests on a WWIV-map with busy waters (so one of the hardest stress tests available with vdips set of variants) showed a significant improvement (non-responsive to runtimes of under 100 ms).

For the time being the changes can be found here:
https://github.com/Sleepcap/vDiplomacy/commit/1686d60b58eec754a3b6ddcbb02b0dd52ae0a88a#diff-c2d37d311b98764477115723be296471

Issue #178, opened 2015-08-31
Javascript issue on the world map


TimothyJones:
When there are a lot of possible convoy routes (eg, when all the seas are full on the world map), the board can sometimes kill the browser. From a member:

> In one of my games, I don't seem to be able to give move orders requiring a convoy. (gameID=162651)
> 
> e.g. if I try to order South Australia to Thailand, I get a series of "Warning: Unresponsive script" errors in Firefox:
> 
> A) "Script: http://webdiplomacy.net/contrib/js/prototype.js:763"
> 
> Then if I continue (rather than kill) each one, I get (infinitely?) more such errors:
> 
> B) "Script: http://webdiplomacy.net/javascript/board/model.js?ver=4:59"
> C) "Script: http://webdiplomacy.net/contrib/js/prototype.js:703"
> D) "Script: http://webdiplomacy.net/javascript/board/model.js?ver=4:326"
> I don't know if there is ever a point where this will end, but it's very tedious, so I've not checked further.
>  If I cancel at that point and try to save the move, I get this from your page:
> "Parameter 'toTerrID' set to invalid value '133'"

TimothyJones:
@Sleepcap - is this fixed on vDiplomacy? I had a peek through the javascript files you've modified, but I couldn't see anything obvious that was a fix for this.

Sleepcap:
It does not do the validation of these largemaps in the javascript-code.
It's not really a fix, more a "duct tape"...

PR #600, opened 2019-08-11
Better convoy code from vdip

diptobi1:
This commit replaces the original WebDip convoy code by an alternative version I made for vdip one and a half year ago. It solves issues with the original code leading to unresponsive behavior on large maps like World with many fleets on Sea territories and therefore many possible convoy paths. This behavior is documented in the issues #178 and #189.

As this is active code on vdip for quite some time now and the last issues were patched out over a half year ago, the code can now be considered stable and might therefore be a good candidate for porting to webdip.

The new code basically replaces the old rather simple but inefficient path searching code by a much more sophisticated one specialized on the problem of finding a path A-B-C without any crossings between the sub-paths A-B and B-C. Unfortunately, the code blow up a bit compared to the original version and is probably hard to review. But the complexness turned out to be necessary to handle the extreme cases of vdip's large variants. After all, this is a problem of exponential complexity, so every small negligence that comes with simplified code leads to a serious worsening of runtime.

I cherry picked all commits made to the code in the vdip repository to make sure, the changes are complete and to preserve the development history in case this is needed for future development. Hence the high number of commits in this pull request.

While nearly everything of this code is original vdip code I made one additional change to include issue #227, too:
Commit 6c05343 adds an additional check for the convoy-from-choices very similar to the check for support-move-from-choices. The latter was also part of the original webdip code, though deactivated due to efficiency issues with the original convoy code. I activated the support move check when I introduced the new convoy code on vdip. Though I was not aware until reading issue #227 that the same check was missing for the convoy-from-choices. But as both code snippets are rather similar, short and understandable, I hope that tests on my local server are enough for the new change.

Overall the successful porting was tested on my local installation on a map of World Diplomacy with fleets placed on every sea territory. In contrast to the original code, this scenario does not freeze the browser when setting convoys with the new code in place. Instead checks are performed apparently immediately (execution runtimes in the ms range). Functionality for setting convoys and support moves for the several roles a fleet can take in that problem have been tested in this scenario.

Let me know, if there are any questions.


jmo1121109:
Okay I had squigs review the new code. I'm confident in the code that's been on vDip for a while. So I'll give this a review myself tomorrow as a final sanity check and then I'll work on merging it in. Thanks for this. Can you incriminate the definitions table, we had a PR recently merged that changed the JS version, so that's a merge conflict right now.

diptobi1:
Alright. Conflict is fixed.

Issue #49, opened 2014-11-17
Stop banned users from having their points returned on CD takeover


TimothyJones:
As per JMO's report, if a user is banned and then has their Civil Disorder taken over, their points are returned to them. This is probably not the right behaviour.

Expected behaviour: 
- Points you have in a game stay in the game pot when you CD
- When your CD is taken over, your points stay in the pot and are not returned.

Current (unconfirmed) behaviour:
- Points you have in a game stay in the game pot when you CD
- But, points are also returned to you when your CD is taken over.

I suspect this affects more than just banned users.

PR #606, opened 2019-08-15
Adjust Points for Bots

jmo1121109:
Adjust points so that bots will never have points taken or awarded. Keep the calls to the user function that controls point distribution because those functions control the member and game table records being updated for bet and pot. So a bot's "bet" in a game will always be 5 points and when they leave a game it will always remove 5 points from the game pot. This is because games with bots playing will have to be unranked, which will be forced in game creation options in another PR.


jmo1121109:
Fixes https://github.com/kestasjk/webDiplomacy/issues/49

Issue #554, opened 2019-07-11
Member list on world games should render on multiple lines


shdant113:
Example of world member list:

![image](https://user-images.githubusercontent.com/45294386/61020575-ddd00c80-a363-11e9-894d-8dd6b3055396.png)

Example of chaos member list:

![image](https://user-images.githubusercontent.com/45294386/61020605-f2140980-a363-11e9-8031-d1d0b3b2f437.png)

The world member list is cluttered and could easily be spread out across multiple lines to give each country equal, readable space.

PR #608, opened 2019-08-17
Fix issue 554

shdant113:
Any variant >7 players will print on multiple lines on home page


jmo1121109:
Fixed https://github.com/kestasjk/webDiplomacy/issues/554

PR #613, opened 2019-08-19
Game listings on home page will display variant name

shdant113:


jmo1121109:
I sent you a screenshot on discord. This ends up really weird on mobile devices, and with the 50% cutoff it runs too long and adds extra lines. I think if you want to save space, removing the RR % display might be one route since that information isn't useful for the home panel instead of the 50% cutoff.

PR #615, opened 2019-08-23
Game must be paused to be reprocessed

ppaquette:
Fixed a race condition.

During a reprocess:

1) Orders for the current phase are cleared
2) The bot is asked to submit orders for the current phase (since they have been cleared)
3) The units are then moved around, and the phase is changed
4) The orders from the bot are received, but the units are not in the expected position -- or the phase is not the expected phase

This fix the race condition by forcing games to be paused before being reprocessed, therefore the bot will not be asked to submit orders when they are cleared just before being reprocessed.

### DIFF
--- a/admin/adminActionsRestricted.php
+++ b/admin/adminActionsRestricted.php
@@ -678,10 +678,16 @@ class adminActionsRestricted extends adminActionsForum
 		$oldPhase = $Game->phase;
 		$oldTurn = $Game->turn;
 
-		// - Check that the game is still active and can be turned back
+		// - Check that the game can be turned back (i.e. has at least one turn processed)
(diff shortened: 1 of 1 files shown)


jmo1121109:
Is this no longer needed?

ppaquette:
Every time an active game is moved forward or backward manually, the bot logs an error (unexpected phase).

I'm assuming this won't happen that often when the bot goes live, so it's not strictly necessary.

I was just trying to figure out if there were any actual bugs left to fix from the errors in the log, but all of them seem to be 1) server timeout (probably when apache gets restarted), and 2) unexpected phase (increased or decreased by 1).

PR #616, opened 2019-08-23
API - Only returned powers who are still playing

ppaquette:
The API indicates that defeated powers are missing orders, so the bot needs to retrieve the game to figure out that the power has no orderable locations.

This PR fixes the issue by only returning 'Playing' powers.


jmo1121109:
Do you also need the orders for anyone who is in the status of "Left" in case someone has gone CD and a Bot is entering orders for them?

ppaquette:
The members in CD would be returned through: api/responses/members_in_cd.php
This is only for api/responses/unordered_countries.php (i.e. countries played by the bot as a regular user)

PR #627, opened 2019-09-14
Deadlock bug

ppaquette:
@kestasjk @jmo1121109

My best guess is that there are deadlocks in the mySQL database, which hangs the server and crashes the site.

I got this in my logs:
```
<b>Warning</b>:  mysqli_query(): (40001/1213): Deadlock found when trying to get lock; try restarting transaction in <b>/var/www/webdiplomacy.net/public_html/objects/database.php</b> on line <b>375</b><br />
<b>Notice</b>:  Deadlock found when trying to get lock; try restarting transaction in <b>/var/www/webdiplomacy.net/public_html/objects/database.php</b> on line <b>377</b><br />
```

Simple fix is to force all adjudication to happen through gamemaster.php, so games are processed sequentially. This should at least avoid processing multiple games at the same time, which is one the likely causes of the deadlock.

Let me know your thoughts.

### DIFF
--- a/ajax.php
+++ b/ajax.php
@@ -97,39 +97,39 @@ elseif( isset($_REQUEST['context']) && isset($_REQUEST['contextKey']) && isset($
 
 		$results = $O->getResults();
 
-		if( $newReady && !$oldReady )
-		{
-			$results['process']='Checked';
-
-			$Game = libVariant::$Variant->Game($O->gameID);
-
-			if( $Game->processStatus!='Crashed' && $Game->attempts > count($Game->Members->ByID)*2 )
-			{
-				$DB->sql_put("COMMIT");
-				require_once(l_r('gamemaster/game.php'));
-				$Game =libVariant::$Variant->processGame($Game->id);
-				$Game->crashed();
-				$DB->sql_put("COMMIT");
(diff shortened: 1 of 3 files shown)


kestasjk:
Jmo and me think we found a query which might have been causing this. The whole of wD_Users was locking, so everyone had to wait for a table lock. Now we are waiting to see if this resolves the issue.

Really the system should request locks in the order they’re used; when the game is processed if I remember right it grabs the member and game rows for that game in an order that prevents deadlocks.

When you say you got that in your logs do you mean it’s appearing in the API code? If so is there a stack trace so we can see one of the places the deadlock is coming from?


Hopefully it won’t come to stopping instant game processing, because without it users will need to wait up to 5 minutes for the game master to process their games, which would be a nuisance for live games.

kestasjk:
Also on the official site I changed several of the new Tournament* innodb and UserConnections

ppaquette:
You are probably right that the query that jmo fixed was probably the issue.

I only have the HTML output resulting from api.php, which doesn't have any stack traces, so I don't have a way to investigate further.

Let's close this for now, and just monitor the performance of the website.

PR #632, opened 2019-09-22
API Refactoring - v2.0

ppaquette:
@kestasjk @jmo1121109 

For your comments.

This is a more substantial API refactoring that fixes most of the current issues.

We should probably wait a couple of weeks before merging this, to make sure everything works properly first. I wanted to do it while I still had the API specification clear in my mind.

The corresponding python PR is at [here PR 96](https://github.com/diplomacy/diplomacy/pull/96).

1) **Allows CD Bot Ordering**

- Added a dropdown on the create game page that allows bots to submit orders when someone is in civil disorder
- Added an admin menu to toggle this setting
- Added a short description that will appear on every game that shows whether the setting is on/off
- If the bot submits an order for someone in CD, a gamemaster message is logged in the global chatbox

A game needs to meet all the following conditions to be eligible for CD ordering:

- It must have the property 'allowBotCDOrdering' == 'Yes'
- It must be active (i.e. not pre-game or finished)
- It must have a member with 'Left' status
- It must not be part of a tournamenet
- It must have the missingPlayerPolicy set to 'Normal'
- It must be on a variantID that is in $apiConfig['variantIDs']
- It must be 'NoPress' or the $apiConfig['noPressOnly'] must be set to false
- It must be on valid gameID if $apiConfig['restrictToGameIDs'] is set

2) **Specification upgraded to support press games**

- Added routes so the bot can play full press games
- The bot can read its own messages, and send messages to other players
- The bot can vote on Draw, Pause, Cancel, Concede
- The bot can toggle the ready flag as necessary
- The bot can query games having new messages after a certain timestamp

3) **Implemented fine-grained permission level**

- Completely rewritten the API permission level
- Now only one permission per API key (canReplaceUsersInCD)
- Every route has a `validateRoute` method to check for permissions

Building a mobile app should also be much easier with this API, since most of the core functionalities are supported (except create_game, find_games, join_game).


jmo1121109:
Oh wow, didn't realize a press version was near. It'll take me some time to look through all this. So your suggestion of a couple weeks is good timing wise lol.

Issue #636, opened 2019-10-05
Feature Request: Live-To-Postal


Restitution-Dip:
An issue with live games is that they last an extremely indeterminate amount of time; a 5-minute turns game can last any given player between 45 minutes and 5 hours if it's played to completion.

A result of this seems to be that Live games end early, or have people drop out or throw the game when they get tired.

A way to fix this would be to have two additional settings available for live games: "transition after" and "next phase time".

Basically, after "transition after" amount of time passes, the next phase will complete, and then phase times will switch to the "next phase time" value.

I think this feature would significantly improve the live game experience, as I think a lot of players would enjoy playing the most interesting part of the game live, and then finishing off the stalemate lines and grinding over the next week or two.

If there's any interest in this feature, I could give a shot at implementing it (though I am by no means a web dev).

jmo1121109:
This is on my eventual to-do list, if it is something you would like to tackle you're more than welcome. My time is somewhat limited right now, but I'd be around for questions.

Restitution-Dip:
Cool, I think I will give it a shot.

PR #637, opened 2019-10-22
Automatic phase transitions (Issue 636)

Restitution-Dip:
Closes: #636 

Frontend and backend work for automatic phase transitions for live games.

### DIFF
--- a/gamecreate.php
+++ b/gamecreate.php
@@ -98,17 +98,20 @@ if( isset($_REQUEST['newGame']) and is_array($_REQUEST['newGame']) )
 		}
 
 		$input['nextPhaseMinutes'] = (int)$input['nextPhaseMinutes'];
+		# Case when nextPhaseMinutes is an invalid value.
 		if ( $input['nextPhaseMinutes'] < 5 or $input['nextPhaseMinutes'] > 1440*10 or $input['nextPhaseMinutes'] < $input['phaseMinutes'])
 		{
 			throw new Exception(l_t("The next phase value is too large or small; it must be between 5 minutes and 10 days, and must be larger or equal to the value of the original phase value."));
 		}
 
-		if ($input['phaseMinutes'] > 30){
+		# Case when the game is not a live game to begin with.
+		if ($input['phaseMinutes'] > 60){
 			$input['nextPhaseMinutes'] = $input['phaseMinutes'];
 		}
 
 		$input['phaseSwitchPeriod'] = (int)$input['phaseSwitchPeriod'];
 
+		# If we're not doing a live phase transition, then we don't need a phase switch period.
 		if ($input['phaseMinutes'] == $input['nextPhaseMinutes']){
 			$input['phaseSwitchPeriod'] = -1;
 		}
--- a/gamecreate.php
+++ b/gamecreate.php
@@ -98,13 +98,13 @@ if( isset($_REQUEST['newGame']) and is_array($_REQUEST['newGame']) )
 		}
 
 		$input['nextPhaseMinutes'] = (int)$input['nextPhaseMinutes'];
-		if ( $input['nextPhaseMinutes'] < 5 or $input['nextPhaseMinutes'] > 1440*10 or $input['nextPhaseMinutes'] < $input['phaseMinutes']))
+		if ( $input['nextPhaseMinutes'] < 5 or $input['nextPhaseMinutes'] > 1440*10 or $input['nextPhaseMinutes'] < $input['phaseMinutes'])
 		{
 			throw new Exception(l_t("The next phase value is too large or small; it must be between 5 minutes and 10 days, and must be larger or equal to the value of the original phase value."));
 		}
 
 		if ($input['phaseMinutes'] > 30){
-			$input['nextPhaseMinutes'] = $input['phaseMinutes']];
+			$input['nextPhaseMinutes'] = $input['phaseMinutes'];
 		}
 
 		$input['phaseSwitchPeriod'] = (int)$input['phaseSwitchPeriod'];
--- a/gamecreate.php
+++ b/gamecreate.php
@@ -45,7 +45,7 @@ if( isset($_REQUEST['newGame']) and is_array($_REQUEST['newGame']) )
 		$form = $_REQUEST['newGame']; // This makes $form look harmless when it is unsanitized; the parameters must all be sanitized
 
 		$input = array();
-		$required = array('variantID', 'name', 'password', 'passwordcheck', 'bet', 'potType', 'phaseMinutes', 'joinPeriod', 'anon', 'pressType', 'missingPlayerPolicy','drawType','minimumReliabilityRating','excusedMissedTurns');
+		$required = array('variantID', 'name', 'password', 'passwordcheck', 'bet', 'potType', 'phaseMinutes', 'nextPhaseMinutes', 'phaseSwitchPeriod', 'joinPeriod', 'anon', 'pressType', 'missingPlayerPolicy','drawType','minimumReliabilityRating','excusedMissedTurns');
 
 		$playerTypes = 'Members';
 
@@ -96,6 +96,23 @@ if( isset($_REQUEST['newGame']) and is_array($_REQUEST['newGame']) )
 		{
 			throw new Exception(l_t("The phase value is too large or small; it must be between 5 minutes and 10 days."));
 		}
+
+		$input['nextPhaseMinutes'] = (int)$input['nextPhaseMinutes'];
+		if ( $input['nextPhaseMinutes'] < 5 or $input['nextPhaseMinutes'] > 1440*10 or $input['nextPhaseMinutes'] < $input['phaseMinutes']))
+		{
+			throw new Exception(l_t("The next phase value is too large or small; it must be between 5 minutes and 10 days, and must be larger or equal to the value of the original phase value."));
+		}
+
+		if ($input['phaseMinutes'] > 30){
+			$input['nextPhaseMinutes'] = $input['phaseMinutes']];
+		}
+
+		$input['phaseSwitchPeriod'] = (int)$input['phaseSwitchPeriod'];
+
+		if ($input['phaseMinutes'] == $input['nextPhaseMinutes']){
+			$input['phaseSwitchPeriod'] = -1;
+		}
+
 
 		$input['joinPeriod'] = (int)$input['joinPeriod'];
 		if ( $input['joinPeriod'] < 5 or $input['joinPeriod'] > 1440*14 )
@@ -190,6 +207,8 @@ if( isset($_REQUEST['newGame']) and is_array($_REQUEST['newGame']) )
 			$input['bet'], 
 			$input['potType'], 
 			$input['phaseMinutes'], 
+			$input['nextPhaseMinutes'],
+			$input['phaseSwitchPeriod'],
 			$input['joinPeriod'], 
 			$input['anon'], 
 			$input['pressType'], 
--- a/gamepanel/game.php
+++ b/gamepanel/game.php
@@ -193,43 +193,6 @@ class panelGame extends Game
 
 		return $buf;
 	}
-
-	/*
-	function phaseSwitchInfo()
-	{
-		$buf = '';
-		
-		if ($this->phase == 'Finished' or $this->phaseSwitchPeriod <= 0){
-			return $buf;
-		}
-		
-		$buf .= '<div class="titleBarRightSide">'.
-					l_t('Next phase length: <span class="nextPhaseMinutes"><strong>%s</strong> /phase</span>', libTime::timeLengthText($this->nextPhaseMinutes * 60)).'</div>';
-			
-		if ($this->startTime > 0){
-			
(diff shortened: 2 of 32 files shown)


Restitution-Dip:
I've finished this up, and tested it fairly extensively.

jmo1121109:
Hey, sorry this took so long to get to, I'm starting a syntax review now and then I'll do a follow up with any functionality problems I find. Obviously this took forever for us to get to, so zero rush on any of this.

locales/English/gamecreate.php
```diff
@@ -66,15 +66,52 @@
...
 
 				foreach ($phaseList as $i) { print '<option value="'.$i.'"'.($i==1440 ? ' selected' : '').'>'.libTime::timeLengthText($i*60).'</option>'; }
 			?>
 			</select>
+			
+			<p id="phaseSwitchPeriodPara">
+				<strong>Time Until Phase Swap</strong></br>
+				<select class = "gameCreate" id="selectPhaseSwitchPeriod" name="newGame[phaseSwitchPeriod]">
+				<?php
+				$phaseList = array(-1, 10, 15, 20, 30, 60, 120, 240, 360, 480, 600, 720, 840, 960, 1080, 1200, 1320,
```
jmo1121109: I think there's 2 ways this can be improved. 1 I think is needed, the other is just a nice to have. 

1. (needed) only give offerings from 10 minutes - 6 hours max, and have every hour be an option after an hour, so like 
-1, 10,15,20,30,60,120,180,240,300,360. Anything above that seems too long for the live game scene. 
2. (ideal but not needed) adjust offerings allowed based on the phase length, so don't allow a shorter time to switch than the phase. Cause if the phase length is 30 minutes, but it switches after 10 minutes then it doesn't make sense.

javascript/timeHandler.js
```diff
@@ -152,7 +152,7 @@ function updateTimers() {
 	
 	var timeFrom = Math.floor((new Date).getTime() / 1000);
 	
-	$$(".gameTimeRemaining .timeremaining").map(function(c) {
+	$$(".timeremaining").map(function(c) {
```
jmo1121109: What is the need for removing the .gameTimeRemaining? The reason I ask is because anytime we have to change the js files it can cause problems. We do somewhat address this by having a file, global\definitions.php, that has a js version in it to help force cache busting, but some phone and browser configurations still keep the js files cached, and having the wrong version will throw errors for people. If this change is needed, include the definitions.php file with the js version incremented. As a tip too, we do the same thing with css so if you ever have to change css, you can cache bust it in the same file.

install/FullInstall/fullInstall.sql
```diff
@@ -902,3 +902,23 @@ UPDATE `wD_Misc` SET `value` = '161' WHERE `name` = 'Version';
 
 ALTER TABLE `wD_Games` ADD COLUMN `playerTypes` enum('Members', 'Mixed', 'MemberVsBots') DEFAULT 'Members' NOT NULL;
 ALTER TABLE `wD_Backup_Games` ADD COLUMN `playerTypes` enum('Members', 'Mixed', 'MemberVsBots') DEFAULT 'Members' NOT NULL;
+
+ALTER TABLE `wD_Games`
+ADD COLUMN `nextPhaseMinutes` int(10) UNSIGNED DEFAULT 0 NOT NULL AFTER `phaseMinutes`,
+ADD COLUMN `phaseSwitchPeriod` int(10) DEFAULT -1 AFTER `nextPhaseMinutes`,
+ADD COLUMN `createTime` int(10) UNSIGNED DEFAULT NULL,
```
jmo1121109: I can't find where this new column createTime is actually used. From what I can see it isn't being utilized anywhere in the code aside from being loaded in the game object.

gamepanel/game.php
```diff
@@ -194,6 +194,72 @@ function gameIcons()
 		return $buf;
 	}
 
+	/*
```
jmo1121109: We'll want to remove any commented out code that isn't needed for the new functionality.

gamepanel/game.php
```diff
@@ -194,6 +194,72 @@ function gameIcons()
...
+						l_t('Phase switch: <strong>%s</strong> after game start', $timeTillNextPhase).'</div>';	
+		}
+								
+
+		return $buf;
+	}
+	*/
+	
+	function phaseSwitchInfo()
+		{
```
jmo1121109: This isn't a huge deal but just from a readability standpoint, we try and keep the indentation of { at the same as the function definition line. 
so

function phaseSwitchInfo()
{

gamepanel/game.php
```diff
@@ -194,6 +194,72 @@ function gameIcons()
...
+		$buf = '';
+		
+		if ($this->phase == 'Finished' or $this->phaseSwitchPeriod <= 0){
```
jmo1121109: For anything in php the norm in our codebase is 
if X 
{
     indented code
}

and anything written in js gets the format 

if X {
    indented code
}

gamecreate.php
```diff
@@ -97,6 +97,23 @@
 			throw new Exception(l_t("The phase value is too large or small; it must be between 5 minutes and 10 days."));
 		}
 
+		$input['nextPhaseMinutes'] = (int)$input['nextPhaseMinutes'];
+		if ( $input['nextPhaseMinutes'] < 5 or $input['nextPhaseMinutes'] > 1440*10 or $input['nextPhaseMinutes'] < $input['phaseMinutes'])
+		{
+			throw new Exception(l_t("The next phase value is too large or small; it must be between 5 minutes and 10 days, and must be larger or equal to the value of the original phase value."));
+		}
+
+		if ($input['phaseMinutes'] > 30){
```
jmo1121109: Could you just add some comments in here explaining why these different logic blocks exist? I'm pretty confident I get them but for any other devs or vDip owners, etc who pull this down having it commented with why is super useful.

gamepanel/game.php
```diff
@@ -203,26 +269,37 @@ function titleBar()
...
+			
+
+				
+		$rightMiddle .= '</div>';
+		
+		$rightBottom = '<div class="titleBarRightSide">'.
+					l_t('%s excused missed turn','<span class="excusedNMRs">'.$this->excusedMissedTurns.'</span>
+					').
+				'</div>';
+		// $rightBottom .= $this->phaseSwitchInfo();
```
jmo1121109: I think this comment can be removed as well.

gamepanel/game.php
```diff
@@ -231,21 +308,28 @@ function titleBar()
...
-			'.$rightBottom.'
+			'.$rightMiddle.'
 			'.$leftBottom.'
 			<div style="clear:both"></div>
-			';
+			'.$rightBottom.'
+			<div style="clear:both"></div>';
 
-		$buf .= $this->gameVariants();
+		// $buf .= $this->phaseSwitchInfo();
```
jmo1121109: Same with these comments

install/1.61-1.62/update.sql
```diff
@@ -0,0 +1,20 @@
...
+ADD COLUMN `phaseSwitchPeriod` int(10) DEFAULT -1 AFTER `nextPhaseMinutes`,
+ADD COLUMN `createTime` int(10) UNSIGNED DEFAULT NULL,
+ADD COLUMN `startTime` int(10) UNSIGNED DEFAULT NULL;
+
+UPDATE wD_Games
+SET nextPhaseMinutes = phaseMinutes
+WHERE nextPhaseMinutes = 0;
+
+
+ALTER TABLE `wD_Backup_Games`
```
jmo1121109: Good catch on the backup tables, they're often missed

locales/English/gamecreate.php
```diff
@@ -347,4 +384,45 @@ function setBotFill(){
...
+selectPhaseSwitchPeriod = document.getElementById("selectPhaseSwitchPeriod");
+phaseSwitchPeriodPara = document.getElementById("phaseSwitchPeriodPara");
+
+selectPhaseMinutes = document.getElementById("selectPhaseMinutes");
+
+nextPhaseMinutesPara.style.display = "none";
+phaseSwitchPeriodPara.style.display = "none";
+
+
+const updatePhasePeriod = () => {
```
jmo1121109: Unfortunately some of the arrow function logic here isn't going to work correctly on older phones and browsers, we don't have a back compiler to downgrade from es6 to lower versions so things like arrow functions and const/let don't function. Would you mind rewriting these with base js functionality?

locales/English/gamecreate.php
```diff
@@ -347,4 +384,45 @@ function setBotFill(){
...
+	console.log('called!!!');
+	console.log(selectPhaseMinutes.value);
```
jmo1121109: Let's pull out the console.log statements.

locales/English/gamecreate.php
Restitution-Dip: I did 1, but haven't found time for 2. Might find that time in future.

javascript/timeHandler.js
Restitution-Dip: IIRC, this is because there is a .timeremaining which is not a child of .gameTimeRemaining which also needs to be decrementing correctly. (I believe it's the timer until the phase swaps).

Restitution-Dip: I've incremented the version.

gamecreate.php
Restitution-Dip: Added comments.

install/1.61-1.62/update.sql
Restitution-Dip: Thanks!

gamepanel/game.php
Restitution-Dip: Done

gamepanel/game.php
Restitution-Dip: Done

gamepanel/game.php
Restitution-Dip: Done

gamepanel/game.php
Restitution-Dip: Donezo!

gamepanel/game.php
Restitution-Dip: Fixed

install/FullInstall/fullInstall.sql
Restitution-Dip: Removed all references to createTime

locales/English/gamecreate.php
Restitution-Dip: Done.

locales/English/gamecreate.php
Restitution-Dip: Done

locales/English/gamecreate.php
jmo1121109: Okay gotcha, I'll probably do part 2 before I merge it into the code base. Thanks for doing 1.

install/FullInstall/fullInstall.sql
jmo1121109: Thanks!

gamepanel/game.php
jmo1121109: Thanks!

install/1.61-1.62/update.sql
```diff
@@ -0,0 +1,18 @@
+ALTER TABLE `wD_Games`
+ADD COLUMN `nextPhaseMinutes` int(10) UNSIGNED DEFAULT 0 NOT NULL AFTER `phaseMinutes`,
+ADD COLUMN `phaseSwitchPeriod` int(10) DEFAULT -1 AFTER `nextPhaseMinutes`,
+ADD COLUMN `startTime` int(10) UNSIGNED DEFAULT NULL;
+
+UPDATE wD_Games
+SET nextPhaseMinutes = phaseMinutes
+WHERE nextPhaseMinutes = 0;
+
```
jmo1121109: You need to add an update statement to this file to move the version of the sql file up to match the version in the definitions file otherwise the site code version and sql version won't match upon go live and the site will crash.

install/FullInstall/fullInstall.sql
```diff
@@ -902,3 +902,21 @@ UPDATE `wD_Misc` SET `value` = '161' WHERE `name` = 'Version';
...
+WHERE `nextPhaseMinutes` = 0;
+
+ALTER TABLE `wD_Backup_Games`
+ADD COLUMN `nextPhaseMinutes` int(10) UNSIGNED DEFAULT 0 NOT NULL AFTER `phaseMinutes`,
+ADD COLUMN `phaseSwitchPeriod` int(10) DEFAULT -1 AFTER `nextPhaseMinutes`,
+ADD COLUMN `startTime` int(10) UNSIGNED DEFAULT NULL;
+
+UPDATE `wD_Backup_Games`
+SET `nextPhaseMinutes` = `phaseMinutes`
+WHERE `nextPhaseMinutes` = 0;
```
jmo1121109: Same sql update for version needed here.

Restitution-Dip: Done

install/1.61-1.62/update.sql
Restitution-Dip: Done

locales/English/gamecreate.php
Restitution-Dip: I ended up doing #2, it's pushed.

PR #643, opened 2019-12-06
fixed dark mode chat colours

firexe:
Like brought up by Don in forum, the colours for Italy and Egypt in Modern were swapped in dark mode, when compared to the normal version.


shdant113:
Good catch, thank you for this.

jmo1121109:
As a quick note, when you do css changes, you need to increment the css version in the file global/definitions.php or it won't show for most people. I'll cover that for this though.

PR #653, opened 2020-03-03
Fixes for getting started doing development

splittingred:
* Updates the README to be clear about the PHP version support
* Updates the README to suggest using FullInstall as opposed to each mysql file during setup
* Fixes constructors that are named as the class (this causes PHP deprecation warnings)
* Fixes uninitialized variable in processUserForm.php


splittingred:
@jmo1121109 @shdant113 Can I have a review?

jmo1121109 (approved):
Thank you for these!

PR #654, opened 2020-03-03
Fix assertion check to be PHP7+ compatible, fixes #381

### LINKED ISSUE #381: PHP 7.X =>Error triggered: assert(): Calling assert() with a string argument is deprecated
Changing to PHP 7.X (X=1, 2 or 3) will throw an error on the "Games tab".
---
Error triggered: assert(): Calling assert() with a string argument is deprecated

Error: "assert(): Calling assert() with a string argument is deprecated"
Raised: "/usr/www/users/vdiplo/vDiplomacy/gamesearch/searchItems.php"
Line: "271"
---
As PHP 7.0 is really old now webhosters are removing support for this version. Might be a good time to work on these funktions to make them future-proof.

splittingred:
> 7.0.0 | assert() is now a language construct and not a function. assertion can now be an expression. The second parameter is now interpreted either as an exception (if a Throwable object is given), or as the description supported from PHP 5.4.8 onwards.

https://www.php.net/manual/en/function.assert.php

Also fixes some reference errors with Countable that show up in PHP 7.2+.

### DIFF
--- a/map/drawMap.php
+++ b/map/drawMap.php
@@ -341,14 +341,18 @@ abstract class drawMap
 	{
 		foreach($this->orderArrows as &$orderArrow)
 		{
-			foreach($orderArrow as $name => &$param) {
+			foreach($orderArrow as $name => &$param)
+			{
 				if ($name == 'headAngle') $param = M_PI / $param;
 
 				if (!is_array($param)) continue;
 
-				if (count($param) == 3) {
+				if (count($param) == 3)
+				{
 					$param = $this->color($param);
-				} else if (count($param) == 2) {
+				}
+				else if (count($param) == 2)
+				{
 					$param = $param[$this->smallmap ? 0 : 1];
 				}
 			}
--- a/map/drawMap.php
+++ b/map/drawMap.php
@@ -341,17 +341,16 @@ abstract class drawMap
 	{
 		foreach($this->orderArrows as &$orderArrow)
 		{
-			foreach($orderArrow as $name => &$param)
-			{
-				if ( $name == 'headAngle' ) $param = M_PI/$param;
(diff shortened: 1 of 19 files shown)


jmo1121109:
@Sleepcap can you take a look at this and see if this solves the issues you were having with php higher than 7 on vDip?

diptobi1:
This does not fix all Countable reference errors yet, does it? (https://github.com/kestasjk/webDiplomacy/issues/656)

splittingred:
@diptobi1 No, not all of them. That's a larger PR.

splittingred:
Honestly the best approach here would be to merge #655, then add [phpcs](https://github.com/squizlabs/PHP_CodeSniffer) + [PHPCompatibility](https://github.com/PHPCompatibility/PHPCompatibility) to dev-require to check for PHP 7.2 support automatically.

splittingred:
@jmo1121109 @shdant113 Can I have a review?

map/drawMap.php
```diff
@@ -341,17 +341,16 @@ protected function loadOrderArrows()
 	{
 		foreach($this->orderArrows as &$orderArrow)
 		{
-			foreach($orderArrow as $name => &$param)
-			{
-				if ( $name == 'headAngle' ) $param = M_PI/$param;
+			foreach($orderArrow as $name => &$param) {
```
shdant113, line 344: Can each of the opening and closing brackets within this loop live on their own line to conform with rest of the style throughout the codebase?

splittingred:
@shdant113 Done, thanks!

jmo1121109:
@kestasjk ended up taking a number of these, and then making the rest of the needed changes for php 7+ when he upgraded the server, so closing this one out.

PR #655, opened 2020-03-03
Use composer for dependency management

splittingred:
### Proposal: Add Composer for dependency management

Modern PHP apps utilize composer for dependency management to allow for external library usage, [PSR-4](https://www.php-fig.org/psr/psr-4/) standards compatibility, and built-in autoloading.

This adds composer to webDip, which only adds one step to be run during initial install (`composer install`). Note: this would need to be run whenever the code is deployed to the actual server, along with installing composer on it first.

Thoughts? This could allow webDip to benefit from a bunch of external libraries (such as templating, API generation, serialization, DB access, caching, etc).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,4 @@ variants/KnownWorld_901/
 ghostRatingData.*
 gr.php
 adminInfo.php
+vendor/
--- a/README.txt
+++ b/README.txt
@@ -55,6 +55,14 @@ Edit config.sample.php to work with your setup, being very careful to read the w
 about security issues. The salts/secrets, errorlog/orderlog directories, can all
 leave your server wide open if you don't set them right. Rename to config.php when ready.
 
+=> Composer
+
+Install Composer via https://getcomposer.org/. Then run:
+
+    composer install
+
+This will install all external dependencies that webDiplomacy uses.
+
 => Log-on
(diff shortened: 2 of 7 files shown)


diptobi1:
I concur. The external phpMailer library already includes some traces of composer. And the library I would like to use for push notifications has some dependencies itself, too, that could be best managed with composer.

Webdip's software already uses an autoload-function itself for the variant extensions, though. And as far as I can judge, this would be broken by your current pull request (had the very same problem just a few days ago). Better add the original autoload function to the spl autoload queue with spl_autoload_register() in variants/variant.php.

splittingred:
@diptobi1 Good catch - thanks! I've updated the branch to consolidate the PHPMailer autoloader (since we're on PHP 7.0+ now), and adjusted the variant autoloader to use spl_autoload_register with a closure.

TimothyJones:
Awesome work!

splittingred:
@jmo1121109 @shdant113 Thoughts here?

jmo1121109:
Hey, at a first glance this looks really neat. I am not familiar with much in the way of php libraries, and anything being installed on the server has to run through @kestasjk, for security purposes, who is not currently active. So I do want to set the expectation that this may sit for a time.

PR #658, opened 2020-03-21
Fix broken cancel game logging

shdant113:


shdant113:
Updated with https://github.com/kestasjk/webDiplomacy/pull/659.

PR #665, opened 2020-03-30
Add new admin tool

shdant113:
Allows admins to change a player's username, creates table wD_UsernameHistory

### DIFF
--- a/admin/adminActionsRestricted.php
+++ b/admin/adminActionsRestricted.php
@@ -553,20 +553,25 @@ class adminActionsRestricted extends adminActionsForum
 
 		if( !isset($params['reason']) || strlen($params['reason'])==0 )
 		{
-			return "Could not change username because no reason was given.";
+			return l_t("Could not change username because no reason was given.");
 		}
 
 		$changeReason = $DB->msg_escape($params['reason']);
 
 		// check if username exists
-		list($result) = $DB->sql_row("SELECT username FROM wD_Users WHERE id = '".$userID."' AND username = '".$newUsername."'");
+		list($result) = $DB->sql_row("SELECT username FROM wD_Users WHERE username = '".$newUsername."'");
 		if (!empty($result)) 
 		{
-			return "This username has already been taken by another user.";
+			return l_t("This username has already been taken by another user.");
 		}
 		
-		// get and store old username, set new username
+		// get and store old username and if it exists set new username
 		list($oldUsername) = $DB->sql_row("SELECT username FROM wD_Users WHERE id = ".$userID);
+		if (empty($oldUsername))
+		{
+			return l_t("User id %s has not been found.", $userID);
+		}
+
 		$time = time();
 		$changedBy = $User->username;
 
@@ -587,7 +592,7 @@ class adminActionsRestricted extends adminActionsForum
 			);
 		}
 
-		return "This user's username has been changed.";
+		return l_t("This user's username has been changed from %s to %s.",$oldUsername, $newUsername);
 	}
 	
 	public function giveBot(array $params)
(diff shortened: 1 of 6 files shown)


admin/adminActionsRestricted.php
```diff
@@ -537,6 +542,53 @@ public function takeForumModerator(array $params)
...
+		// get and store old username, set new username
+		list($oldUsername) = $DB->sql_row("SELECT username FROM wD_Users WHERE id = ".$userID);
+		$time = time();
+		$changedBy = $User->username;
+
+		$DB->sql_put(
+			'INSERT INTO wD_UsernameHistory (userID, oldUsername, newUsername, date, reason, changedBy) 
+			VALUES ("'.$userID.'", "'.$oldUsername.'", "'.$newUsername.'", "'.$time.'", "'.$changeReason.'", "'.$changedBy.'")'
+		);
+		$DB->sql_put("UPDATE wD_Users SET username = '".$newUsername."' WHERE id = ".$userID);
```
jmo1121109: add a limit 1 to the end of this update just for safety.

admin/adminActionsRestricted.php
```diff
@@ -537,6 +542,53 @@ public function takeForumModerator(array $params)
...
+
+		if( !isset($params['reason']) || strlen($params['reason'])==0 )
+		{
+			return "Could not change username because no reason was given.";
+		}
+
+		$changeReason = $DB->msg_escape($params['reason']);
+
+		// check if username exists
+		list($result) = $DB->sql_row("SELECT username FROM wD_Users WHERE id = '".$userID."' AND username = '".$newUsername."'");
```
jmo1121109: This won't work, you're checking on id and username, you need to check only on username otherwise you won't find if someone else already has that username.

admin/adminActionsRestricted.php
```diff
@@ -537,6 +542,53 @@ public function takeForumModerator(array $params)
...
+		if (!empty($result)) 
+		{
+			return "This username has already been taken by another user.";
+		}
+		
+		// get and store old username, set new username
+		list($oldUsername) = $DB->sql_row("SELECT username FROM wD_Users WHERE id = ".$userID);
```
jmo1121109: Can you add a check here to make sure that an old username exists, it'll serve as a check to make sure a valid userID was given.

admin/adminActionsRestricted.php
```diff
@@ -537,6 +542,53 @@ public function takeForumModerator(array $params)
...
+		{
+			$newUsernameClean = strtolower($newUsername);
+
+			$DB->sql_put(
+				"UPDATE phpbb_users SET username = '".$newUsername."', username_clean = '".$newUsernameClean."' 
+				WHERE username = '".$oldUsername."' AND webdip_user_id = '".$userID."' limit 1"
+			);
+		}
+
+		return "This user's username has been changed.";
```
jmo1121109: Can you return "UserID X's username has been updated from Y to Z so they can see if they messed up?

install/FullInstall/fullInstall.sql
```diff
@@ -913,4 +913,13 @@ UPDATE `wD_Backup_Games`
...
 WHERE `nextPhaseMinutes` = 0;
 
-UPDATE `wD_Misc` SET `value` = '162' WHERE `name` = 'Version';
+CREATE TABLE `wD_UsernameHistory` (
+  `userID` mediumint(8) NOT NULL,
+  `oldUsername` varchar(30) NOT NULL,
+  `newUsername` varchar(30) NOT NULL,
+  `date` int(10) unsigned NOT NULL,
+  `reason` varchar(50) NOT NULL,
+  `changedBy` varchat(30) NOT NULL
```
jmo1121109: varchar not varchat :)

PR #668, opened 2020-04-03
New admin tool

shdant113:
Creates tool to allow admins to change registered emails and creates table to store changed emails for moderator case reference

### DIFF
--- a/admin/adminActionsRestricted.php
+++ b/admin/adminActionsRestricted.php
@@ -76,7 +76,7 @@ class adminActionsRestricted extends adminActionsForum
 			),
 			'changeEmail' => array(
 				'name' => 'Change email',
-				'description' => 'Changes user\'s current registered email to the specified email.',
+				'description' => 'Update user\'s registered email.',
 				'params' => array('userID'=>'User ID', 'email'=>'New Email', 'reason'=>'Reason')
 			),
 			'changeUsername' => array(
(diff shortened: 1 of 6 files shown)


admin/adminActionsRestricted.php
```diff
@@ -74,6 +74,11 @@ public function __construct()
 				'description' => 'Takes forum moderator status from the specified user ID.',
 				'params' => array('userID'=>'Mod User ID'),
 			),
+			'changeEmail' => array(
+				'name' => 'Change email',
+				'description' => 'Changes user\'s current registered email to the specified email.',
```
jmo1121109, line 79: Weird English. Can you update this to "Update the email of the user"?

admin/adminActionsRestricted.php
```diff
@@ -543,6 +548,49 @@ public function takeForumModerator(array $params)
...
+		{
+			return l_t("User id %s has not been found.", $userID);
+		}
+
+		$time = time();
+		$changedBy = $User->username;
+
+		$DB->sql_put(
+			'INSERT INTO wD_EmailHistory (userID, oldEmail, newEmail, date, reason, changedBy) 
+			VALUES ("'.$userID.'", "'.$oldEmail['email'].'", "'.$newEmail.'", "'.$time.'", "'.$changeReason.'", "'.$changedBy.'")'
```
jmo1121109, line 586: userId shouldn't be a string in the table so remove the ' ' around it in the Values section.

admin/adminActionsRestricted.php
```diff
@@ -543,6 +548,49 @@ public function takeForumModerator(array $params)
...
+		);
+		$DB->sql_put("UPDATE wD_Users SET email = '".$newEmail."' WHERE id = '".$userID."' limit 1");
```
jmo1121109, line 588: same thing with userId in the where here, remove the ' '

admin/adminActionsRestricted.php
jmo1121109: Same with date/time too I think since you're making that column as an int in the db table

PR #675, opened 2020-05-08
Mod tool for changing game's RR requirement

shdant113:


admin/adminActions.php
```diff
@@ -1248,6 +1253,28 @@ public function generateRegistrationLink(array $params)
...
+		
+		if (!isset($params['newRating']) || (!preg_match('/^[0-9][0-9]?$|^100$/', $params['newRating'])))
+		{
+			return "Please enter a valid reliability rating requirement (0-100).";
+		}
+
+		$gameID = (int)$params['gameID'];
+		$newRating = (int)$params['newRating'];
+
+		$DB->sql_put("UPDATE wD_Games SET minimumReliabilityRating = ".$newRating." WHERE id = ".$gameID);
```
jmo1121109, line 1273: lets throw a limit 1 at the end of that just to be safe

admin/adminActions.php
```diff
@@ -1248,6 +1253,28 @@ public function generateRegistrationLink(array $params)
...
+	public function changeReliability(array $params)
+	{
+		global $DB;
+
+		if (!isset($params['gameID']))
+		{
+			return "Please enter a game ID.";
+		}
+		
+		if (!isset($params['newRating']) || (!preg_match('/^[0-9][0-9]?$|^100$/', $params['newRating'])))
```
jmo1121109, line 1265: Can you run another query that gets the min(rr) from the users in the game and display's an error message if they're trying to make the RR higher than what someone already in the game has?

admin/adminActions.php
```diff
@@ -1248,6 +1253,28 @@ public function generateRegistrationLink(array $params)
...
+		{
+			return "Please enter a valid reliability rating requirement (0-100).";
+		}
+
+		$gameID = (int)$params['gameID'];
+		$newRating = (int)$params['newRating'];
+
+		$DB->sql_put("UPDATE wD_Games SET minimumReliabilityRating = ".$newRating." WHERE id = ".$gameID);
+
+		return "Game ".$gameID."'s reliability rating has been set to ".$newRating;
```
jmo1121109, line 1275: Can you turn this into a link to the game? And in the return message can you pull the old RR too so we can see in the logs what it used to be just in case for any debugging needed?

admin/adminActions.php
shdant113: I was thinking of not doing this because as is a player can play in a game where the min RR is higher than their rating so long as they joined the game before their RR dropped below the minimum. Similarly, it makes sense that a player would be still able to play in a game if the RR requirement was changed if they were already in the game.

jmo1121109: okay

Issue #491, opened 2019-05-19
Import from vDip - Lock Game


jmo1121109:
Ability for mods to freeze votes in games. 

Called Lock/unlock a game.

The goal is to solve mods having to post "Please do not vote to unpause, draw, or cancel this game, doing so will result in docked points" for games that need investigations. This would ideally prevent CD takeovers in this game as well.

PR #676, opened 2020-05-08
Mod tool to remove all draw/unpause/cancel votes

shdant113:


jmo1121109:
@shdant113 can you take a look at this https://github.com/kestasjk/webDiplomacy/issues/491 and see if this can be changed a bit to hit that functionality?

Issue #673, opened 2020-04-11
Wrong colors on Cuba and Heartland


maximeflageole:
# Bug report

Colors of Cuba and Heartland in Fall of american empire are inverted in the chat.

## Expected behaviour

Colors should be right

## Actual behaviour

Dark blue and light blue are inverted in 
the chat. I am not sure if it is the same for the other players. I am B-C (in case it helps)

## Steps to reproduce

Start as B-C. Observe that the colors or Cuba and Heartland are different on the map vs in chat. Be slightly confused and open a bug report on github

jmo1121109:
Thanks for the report, is this in light mode or dark mode?

maximeflageole:
Hi! In dark mode

PR #680, opened 2020-05-23
Heartland and Cuba's colors no longer swapped

adbronico:


adbronico:
https://github.com/kestasjk/webDiplomacy/issues/673

PR #687, opened 2020-06-11
Adding thick borders to the buildable Supply Centers to the Ancient Mediterranean variant

Moxanity:
The Classic variant's map has thicker borders surrounding the original territories, making it easy at a glance to tell which supply centers allow building for their original owner.

This pull request adds this for the Ancient Mediterranean, both the small and large map.
Since this map is much more boat-based, unlike the Classic variant, thicker borders are also added to coast borders (as opposed to just land borders).


jmo1121109:
Hey, sorry for the delay on this. Thank you for making this change!

diptobi1:
I noticed that the new files are not stored as indexed png, causing the map not to be colored correctly (e.g. here: https://webdiplomacy.net/map.php?gameID=318814&turn=10&mapType=large).

Issue #104, opened 2015-02-16
AncMed: able to order army to move without convoy to Baleares 


TimothyJones:
From a user:

>http://screenshooter.net/100368172/jldqdvg

> I am not sure, but shouldn't that be Baleares (convoy)? I mean, I now
it's an extremely small bug, but I almost made a really terrible
decision based on that. Or better, if I put in Baleares would this go
through?

Sleepcap:
This is a javascript-display problem.
It's only accessible "via convoy", and internally it's treated like it's "via convoy", but the code does not append "via convoy" to the string. It's fixed on vDip long ago.

Here is the related thread:
http://forum.webdiplomacy.net/viewtopic.php?f=3&t=569

TimothyJones:
Fantastic info! Thanks. I'll incorporate your fix next time I'm working on it, or if you beat me to it with a pull request that's fine too.

Sleepcap:
How do you want to fix javascript problems?
Editing the JS-code might keep the old code in many users cache...
I made a new version for each fix.
Also I did fix quite a lot of different JS-problems over the last years. It might be worth considering all of them. I'll try to make a list...

TimothyJones:
I was planning to force the browser to update the javascript by putting a version number in the query part of the URL, which is ignored: `whatever.js?version=3`

TimothyJones:
And yes please, a list of possible JS fixes would be excellent!

Sleepcap:
I'm unsure if it's a good idea to put the version-control in the query part.
The JS-files are static and do not change that often, so a forced-reload each time is (IMHO) unneeded.
If you add a version-number to the filename and change this in Orderinterface.php this should be sufficient. 

Currently we have on vDip:
1. Tobis change for the support-move from a coast.
2. My fix for the "via convoy" display issue.
3. All dropdown-menues gets sorted by name (and all convoy orders are sorted at the end)
4. You can't "Save" or "Ready" unfinished orders.
5. If you want to "move" and you can move "via land" or "via convoy" it uses "via land" as default, as this does lead to uncompleted orders many times.

But 4 is a bit redundant, since we implemented 5.

jmo1121109:
@Sleepcap this recently became a problem for us again and unfortunately the thread you mention no longer exists cause Kestas had to archive some old threads a couple years back. Do you have any info on how I can fix this incorrect order being allowed?

PR #688, opened 2020-06-13
Fix for incorrect labelling of convoy choices

diptobi1:
This is a small fix from vDip for issue #104, the incorrect labelling of the via convoy option to the Baleares as normal move.

As you can see in the fix, the issue of the original code was, that for the classification of move choices into local and convoy choices neighboring land territories were always considered directly reachable by an army. The fix replaces the method to get the local choices by the one, that takes the border connections into account.

This fix is from Oli (Sleepcap) that I have extracted from the vdip code base via a diff. Apart from the years of use on vdip I made some basic manual testing the order interface to guarantee that this will work on webdip.

In #104 Oli mentioned some more js features from vdip you might be interested in. The implementations are simple and easy to extract with a diff. Let me know if I should add some of that to the request (support-move fix and 'via land' default choice are already included in webdip).


jmo1121109:
Thank you for this!

PR #692, opened 2020-06-23
GIF Feature

jensforsgard:
Adds the feature to create a GIF-file from the maps from a finished game.

The link to create a GIF-file appears in the Archive bar on the game page. To avoid creation of a large number of GIF-files, the feature is only available for finished games, and the user is not allowed to modify the parameters of the GIF.

- Main code to create the GIF file is a stripped version of the AnimGif class from https://github.com/lunakid/AnimGif, see board/info/gif/animate.php
- The class Game now includes a method create_webDip_animation()
- Added file board/info/gif.php which displays the animation
- board.php and gamepanel has been adjusted accordingly

Possible issues:
The AnimGif class requires the GD extension. Works well in my dev - but not obvious it is available on the server.

Remarks:
The map file "-1-small.map", which shows the starting positions in the game, is deleted from the cache folder once the game is finished, and it is therefor not included in the animation. I believe the animation would be nicer if also this map is included, but that requires that we store copies of maps with starting positions for all variants (including inactive variants) in a separate folder.

### DIFF
--- a/board.php
+++ b/board.php
@@ -121,6 +121,7 @@ if ( isset($_REQUEST['viewArchive']) )
 		case 'Messages': require_once(l_r('board/info/messages.php')); break;
 		case 'Graph': require_once(l_r('board/info/graph.php')); break;
 		case 'Maps': require_once(l_r('board/info/maps.php')); break;
+		case 'Gif': require_once(l_r('board/info/gif.php')); break;
 		case 'Reports':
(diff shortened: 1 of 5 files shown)


jmo1121109:
Just as an fyi, I think this is really cool. I personally don't know enough about the map and adding extensions to the server for this. Kestas will be taking a look at the map in the semi near future so we're gonna have him take a look at this when he is digging into that code area.

Issue #691, opened 2020-06-18
Modern Variant: Large Map Visual Bug


jensforsgard:
On the large map for the Modern variant, there is a quadripoint where Milan, Austria, Croatia, and Venice meet. On the small map, Austria borders Venice but Milan does not border Croatia. The code agrees with the small map.

It is not clear from the variant webpage what the intended border adjacencies are
(see http://www.variantbank.org/results/rules/m/modern2.htm).

The large map on vDip has been adjusted to show Austria bordering Venice.
(see https://www.vdiplomacy.com/variants/Modern2/cache/sampleMapLarge.png)

PR #696, opened 2020-07-23
Modern 2 Venice/Milan Border Fix

PeregrineFlight:
Ports Sleepcap's [2013 vDip fix](https://github.com/Sleepcap/vDiplomacy/commit/1e1d975a2519dd55f0fabeaa631b1ddcf931e01c) to the Modern 2 Large Map to make it clear that Milan does not border Croatia. Fixes issue  #691 .


jmo1121109:
Thanks for this!

PR #702, opened 2020-09-14
Tutorial mode for newly registered users

shdant113:
• Renders modal on home page, game creation form, game listings, and account settings for new users based on cookies expiring after 365 days
• Adds helper functions for modal operations
• libHTML function can be reused for later help modal additions

### DIFF
--- a/index.php
+++ b/index.php
@@ -337,7 +337,7 @@ class libHome
 		if($count==0)
 		{
 			$buf .= '<div class="hr"></div>';
-			$buf .= '<div class="bottomborder"><p class="notice">'.l_t('You\'re not joined to any games!').'<br />
+			$buf .= '<div class="bottomborder"><p class="notice">'.l_t('You have not joined any games!').'<br />
 				'.l_t('Access the <a href="gamelistings.php?tab=">Games</a> '.
 				'link above to find games you can join, or start a '.
 				'<a href="gamecreate.php">New game</a> yourself.</a>').'</p></div>';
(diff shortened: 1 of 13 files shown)


index.php
```diff
@@ -659,7 +659,49 @@ static function forumNewExtern()
...
+	
 	print '<div class="content-bare content-home-header">';// content-follow-on">';
+	
+	// home page tutorial
+	if (isset($_COOKIE['wD-Tutorial-Index'])) 
+	{
+		$tutorialMessage = l_t(
+			'To help you become familiar with webDiplomacy, we prepared a few brief presentations
+			for you that you will find on different pages of the site. You will only see these once apiece, but if you
+			ever want to review what you read, you can check out our <a href="faq.php" target="_blank">Frequently Asked Questions</a>.
```
jmo1121109, line 671: Can we make a button at the top of each page with these with a ? icon that will reset the cookie for that page so someone can re-review these if they want?

javascript/help.js
```diff
@@ -0,0 +1,51 @@
...
+function indexForward(index, tutorial = true) {
+    if (tutorial) {
+        document.getElementsByClassName("tutorial")[index - 1].style["display"] = "none";
+        document.getElementsByClassName("tutorial")[index].style["display"] = "block";
+        document.getElementsByClassName("tutorial-header")[0].scrollIntoView();
+    }
+}
+
+function endTutorial() {
+    document.cookie.split("; ").forEach((c) => {
```
jmo1121109, line 34: Is all this functionality included in base js so it will work on older browsers/phones?

lib/html.php
```diff
@@ -384,6 +384,95 @@ static public function error($message)
...
+	/**
+	 * Renders reusable modal for tutorial mode, future help functionality
+	 * Breaks up message into array of slides based on HTML <br> element
+	 * 
+	 * @param string $page
+	 * @param string $message
+	 */
+	static public function help($page, $message)
+	{
+		if (isset($_COOKIE['wD-Tutorial'])) {
```
jmo1121109, line 396: bracketing inconsistent, next line please for most of the new brackets added in this file.

register/processUserForm.php
```diff
@@ -96,6 +96,14 @@
 
 	$NewUser->options->set($_REQUEST['userForm']);
 	$NewUser->options->load();
+
+	// Give user access to tutorial views for 30 days
```
jmo1121109, line 100: Why just 30 days? Should we have it till they end up checking out the tutorials?

usercp.php
```diff
@@ -139,6 +139,23 @@
 	}
 }
 
+// settings page tutorial
+if (isset($_COOKIE['wD-Tutorial-Settings'])) {
```
jmo1121109, line 143: bracketing on next line for php please

jmo1121109 (changes requested):
Looks awesome, needs a couple tweaks but this should be a huge improvement for our new members.

index.php
jmo1121109: We talked on this and this will be a follow up PR.

register/processUserForm.php
jmo1121109: Thanks for adjusting this

jmo1121109 (approved):

Issue #263, opened 2017-10-06
Error Starting New Game on Fresh Install


jrow3:
# Bug report

Error when trying to start a New Game on a fresh install

Error triggered: file_put_contents(variants/Modern2/cache/territories.js): failed to open stream: No such file or directory

(error edited)

## Expected behaviour

New Game prompt

## Actual behaviour

Error triggered: file_put_contents(variants/Modern2/cache/territories.js): failed to open stream: No such file or directory

Error: "file_put_contents(variants/Modern2/cache/territories.js): failed to open stream: No such file or directory"
Raised: "/home/apog/djrow3.com/diplomacy/variants/install.php"
Line: "89"
userID = 6Variable dump: jsonFileLocation => variants/Modern2/cache/territories.js
mapID => 19
DB => Array: (
putqueries => 11
getqueries => 12
)
territories => Array: (
1 => Array: (

## Steps to reproduce

Continued error.

TimothyJones:
Thanks for the report. I think webdiplomacy doesn't create the cache directories by default. 

Does `variants/Modern2/cache/` exist? If not, you should be able to work around this by creating it.

TimothyJones:
We can fix this by creating the cache directories by default.

jrow3:
I was able to bypass this error by copy and pasting over the Classic/cache folders to Modern2 and Empire4.  It seems the master does not include these files.

TimothyJones:
Thanks! A pull request that contains that change would be welcome if you have the time to make one.

jrow3:
Will do tomorrow

TimothyJones:
I've added those files in commit https://github.com/kestasjk/webDiplomacy/commit/570a630a63946d90b7235397f16dadfd680ae746

TimothyJones:
Yes, we should create the cache directories when they don't exist.

Sleepcap:
These should be moved to the main-cache-directory (cache/variants/VARIANT_SHORTNAME for example)
Storing cached-data in each variant-subdirectory is not really good coding practice.
Maybe use the /lib/cache.php to handle these too...

jmo1121109:
@Sleepcap if this is something you'd feel comfortable doing on vDip I'd happily put the time into porting it back over here. I just don't know how to do it myself.

PR #720, opened 2020-11-13
Patch 1

alexintour:
created cache folders in the variants that were missing it.


alexintour:
this should solve the issue #263

PR #729, opened 2021-05-12
indicate islands' inaccessibility

Nescio0:
Switzerland is impassable and therefore solid black on the map, which is clear and unambiguous. However, Iceland, Ireland, Corsica, Sardinia, Sicily, Crete, Cyprus, and other islands are inaccessible too, yet they aren't indicated as such. Crete and Sardinia are always in the neutral colour, whereas the rest are considered part of the nearest coastal area and change colour accordingly. The current situation is:
* inconsistent: why treat Switzerland, Sardinia, and Crete differently?
* incorrect: Corsica has been French since 1768, not Italian; Iceland was Danish, never British; and Cyprus was a British protectorate since 1878.
* distracting: people look at the map to see what has changed; Iceland changing colour whenever a fleet enters Clyde does not help.

This pull requests proposes treating Switzerland and all islands exactly the same, by using a very dark grey (rgb = (32, 32, 32) or #202020), to indicate units can't move there, and no longer have them change colour, regardless what happens in nearby coastal areas.

This is how it would look at game start:
start_new

For comparison:
webdiplomacy.net:
webdiplomacy
playdiplomacy.com:
playdiplomacy
backstabbr.com:
backstabbr

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,5 @@
 Diplomacy is a popular turn based strategy game in which you battle to control Europe; to win you must be diplomatic and strategic.
+
  
 webDiplomacy lets you play Diplomacy online.
 


Nescio0:
The BuildAnywhere, CustomStart, and FleetRome variants don't have map.png and smallmap.png files, however, ClassicChaos, ClassicFvA, and ClassicGvI do.
Are those actually any different from the maps of the Classic variant?

PeregrineFlight:
Personally, I think the dark grey is too dark. It stands out too much from the rest of the primarily light-coloured map. I've actually done some asking-around in the past on this topic, and most people agree that colouring all impassable spaces black looks bad. 
I did a selection of potential options, which can be found [here](https://drive.google.com/file/d/1W6YKwE0mpe0dudFY96MYKlrscytiIMxM/view?usp=sharing). People on the Discord server seemed to like #4 best, with impassable islands coloured in light grey, but Switzerland left black. (below)
4  Impassable Grey Black Switz - map
4  Impassable Grey Black Switz - smallmap

PR #731, opened 2021-05-12
Gulf of Lion, Marseille, Skagerrak

Nescio0:
This pull request renames three map territories across all variants (yes, I did a `grep` to find all occurrences). The relevant `mapNames.png` and `smallMapNames.png` are updated too.

Gulf of Lyons → Gulf of Lion (it has nothing to do with the city of Lyon)
https://en.wikipedia.org/wiki/Gulf_of_Lion

Marseilles → Marseille (no final `s`)
https://en.wikipedia.org/wiki/Marseille

Skagerrack → Skagerrak (no `c`)
https://en.wikipedia.org/wiki/Skagerrak

PS The relevant commits are:
f9e1d36e53830bf6fcd3f15925f99bf40bc21c51
1c42db367457ea51e956e340a57438039cfecc9e
a169d5e645649bad292dd40bf7ecbb22c2424b32
The other commits were reverting mistakes.
Do a [squash and merge](https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/about-pull-request-merges#squash-and-merge-your-pull-request-commits).


jmo1121109:
Hey, thank you so much for the PR, but I'm going to decline taking it into the codebase. Our spellings intentionally match the board game, and we plan to keep it that way.

Nescio0:
@jmo1121109, in that case I'd like to point out that the names do not exactly match the board game's. Here are the official names from the most recent version (5th edition 2008):
5thEdition2008
Whilst Marseilles does indeed have a final `s`, Gulf of Lyon [sic] does not, nor does Skagerrak have a `c`.
Furthermore, Helgoland Bight is written without an `i`.

Moreover, the spellings are not entirely consistent within webdiplomacy either:
* `variants/classic/install.php` has Gulf of Lyons (with `s`) and Skagerrack (with `c`), as does its `mapNames.png`, however, the `smallmapNames.png` has Gulf of Lyon (without `s`)
* `variants/Modern2/install.php` has Gulf of Lyon (no `s`) and Skagerrak (no `c`), as do its `mapNames.png` and `smallmapNames.png`

PR #744, opened 2022-01-13
Staging -> Staging

hyukcodazen:
Small fix on New Play Beta CTA to redirect user to the Beta map.
Seems like lots of diffs are auto indentation fix.
changes are made on
line 35 ~ 54 on board.php
line 469, 564, 570~580 on members.php
line 526,527 on game.php

### DIFF
--- a/api.php
+++ b/api.php
@@ -19,6 +19,7 @@
  */
 
 define('IN_CODE', 1);
+require_once('header.php');
 require_once('config.php');
 require_once('global/definitions.php');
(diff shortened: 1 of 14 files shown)


kestasjk:
Not sure if you're aware I can see it's redirecting to beta but it doesn't look like the beta board itself has been added?

hyukcodazen:
We have a dummy react app at the moment. Actual beta board is in development and will be added later.

PR #779, opened 2022-05-17
Cold War games forced to unranked 5pt bets

PeregrineFlight:
This adds Cold War to the Game Creation page's code that forces 1v1s to only create unranked, 5pt bet game, to prevent point farming. 
I had forgotten to add this when first adding the Cold War files.


kestasjk:
Thanks

PR #795, opened 2022-06-07
Spectation

adamlerer:
Added support for any player to view any game. I think this preserves the semantics of the old API: basically any information that's available to *all* players in a game (global messages, non-anon order status and draw votes) are available to spectators. Note that this change gives API permission to get state of *any* game to any user with a session key.

image


adamlerer:
Oh, this PR also fixes a potential issue in the message send API: previously, players in no-press games could still *send* messages, they just couldn't *fetch* messages. I don't know of a way to exploit this but it seems bad that players could write messages to the DB even in no-press games.

adamlerer:
Closing because these changes are included in https://github.com/kestasjk/webDiplomacy/pull/797

PR #802, opened 2022-07-01
Release beta v2 0.3

adamlerer:
Release v0.3 of the new beta UI.

IMPORTANT notes:
1. Unlike previous beta versions, this one makes db modifications, meaning we will have to upgrade to 1.70. master is still on 1.68 though, so what do we do? We can either upgrade to 1.70 in prod, or switch the 1.69 and 1.70 changes and switch to 1.69 ...
2. Should this be merged into master or staging?

Feature list:
- turn off "Early Access" guard for Beta UI
- allow games to specify a shorter phase length for R/B phases and grace periods, vs M phases
- Fix handling of civil disorders, pre-game phase, and vote processing in beta UI
- Add blinking for unsaved unit orders
- Add support for beta UI within the webdiplomacy docker configuration
- fix longstanding bug in old UI in rulebook press, where moderator would broadcast a private message to everyone if they sent a message in a R/B phase


adamlerer:
@kestasjk I tried resolving the merge conflicts here, but at least on my private instance, I found that the current master branch isn't working properly (gamemaster doesn't start bot games, cancel doesn't work, etc.) so I was wondering if your changes from a few hours ago might have broken something.

Are changes on master auto-synced to webdiplomacy.net ? Because things are working fine there.

kestasjk:
Hi Adam, 
Hmm that's odd.. 

I changed how votes are handled since it wasn't working reliably; PHP logic detected when a vote passed, and the logic was at a point where I didn't understand the intent. Also because it was in PHP you couldn't search for votes that need processing, you had to detect it when the vote is submitted or game is viewed by a player and then set a flag in memcache for the gamemaster to handle.
I replaced it with a query to search for votes to process that ignores votes from bots entirely. This is a change to the previous logic which did factor in votes from bots, but I couldn't make sense of what it was trying to do where it would sort of allow votes depending on whether the bot had gained or lots supply centers in the last 2-4 turns, so I figured (since it wasn't working for many anyway) I would simplify it right down and see if any issues arose.

Possibly unrelated but FYI: We were also still getting deadlocks / lock timeouts appear in the error logs that I wanted to resolve. When looking into the API I realized it was locking games for update, the most strict and intrusive lock type, to do read-only operations including just checking whether a user was in a game. With hindsight I'm impressed (with the hardware) that the site never became totally inoperable.
Part of the reason is knowing the new board does do a lot of background server requests, and I want to add an instant-play feature so anyone can go to a URL, no account needed, start a bot game and leave when they like, which might increase server load.

Yep webdiplomacy.net syncs with the master branch automatically, and it is in sync at the moment. Is it 1 human games that aren't processing or multi-human games, and are full human games processing? I didn't change how the gamemaster does game processing and it is working so I'm not sure what's going on there. I can see you've got some changes talking about database changes / vote processing etc so maybe there's a merge conflict of some sort. I want to try and finish up a couple of features today but I might be able to look at this merge. I do also want to get staging up to pace with the master branch so you can merge with staging again.

I'll try and bring the database/code version up to 1.71 today, and do a test of a clean install in docker, so you can merge in the database changes neatly into 1.72 via a install/1.71-1.72/update.sql script.

adamlerer:
Thanks @kestasjk . Yeah, the vote logic was hard for me to understand as well, I sort of treated it as a black box. Once you update the code I'll probably start from scratch on how to initiate vote processing from the setVote API call (let me know if you have a suggestion).

Re locking: I didn't write this code, not surprised that this happened because the `getAssociatedGame` function was originally written for a limited purpose and got cargo-culted around for various purposes. Lucky it didn't cause any problems.

What's not working on my end is the processing (starting) of all-bot games. I haven't experimented with anything else.

Thanks, I'll take another pass once you make your changes.

beta-src/src/components/controllers/WDMainController.tsx
```diff
@@ -101,6 +103,14 @@ const WDMainController: React.FC = function ({ children }): React.ReactElement {
             // *not* yet seen (i.e. presumably they saw this phase when they were entering
             // orders for it, before that phase ended and other powers' orders appeared).
             dispatch(gameApiSliceActions.setViewedPhaseToLatestPhaseViewed());
+            if (status.status === "Left") {
```
TimothyJones, line 106: Unrelated to this change, but I think it's better to declare the function outside the JSX inside a `useCallback`, otherwise there may be rendering overhead. In the current form, a new clickhandler function is created every time this component renders.

beta-src/src/components/controllers/WDMainController.tsx
```diff
@@ -79,15 +80,16 @@ const WDMainController: React.FC = function ({ children }): React.ReactElement {
...
     return <Box>Loading...</Box>;
   }
-
   const showOverlay =
-    noPhase || (displayedPhaseKey && overviewKey !== displayedPhaseKey);
+    noPhase ||
+    status.status === "Left" ||
+    (displayedPhaseKey && overviewKey !== displayedPhaseKey);
   if (displayedPhaseKey === null && overview.phase) {
     setDisplayedPhaseKey(overviewKey);
```
TimothyJones, line 88: I might be missing something with this `Controller` pattern, but I believe this call (and the other dispatch calls) have side effects, which means they would properly be placed inside a `useEffect` or an event handler. I don't think it's right to call them directly in the render function.

Personally, I would put this whole check inside a selector - if you can compute it from the main state, you don't need to store it in separate state. Storing things you can compute is risky because you might forget to update them, or you might get out of sync with the main state in other ways.

beta-src/src/components/controllers/WDUnitController.tsx
```diff
@@ -11,26 +8,27 @@ import WDArmyIcon, {
...
-  type,
-  iconState,
+  unit,
+  unitState,
   unitWidth,
   unitHeight,
 }): React.ReactElement {
   const iconWidth =
-    type === "Fleet" ? FLEET_RAW_ICON_WIDTH : ARMY_RAW_ICON_WIDTH;
+    unit.unit.type === "Fleet" ? FLEET_RAW_ICON_WIDTH : ARMY_RAW_ICON_WIDTH;
```
TimothyJones, line 29: Is there a way to write this without needing to access it with `unit.unit`? I feel like there are two different meanings of `unit` here, and it's not obvious what the difference is during a skim read.

As an aside, knowing how to calculate the width and height doesn't feel like the responsibility of a controller - it might read better as a function that returns `{ iconWidth, iconHeight }`

beta-src/src/components/controllers/WDUnitController.tsx
```diff
@@ -49,12 +47,13 @@ const WDUnitController: React.FC<UnitControllerProps> = function ({
       style={{ pointerEvents: "none", overflow: "visible" }}
       transform={transform}
     >
-      {type === "Fleet" && (
-        <WDFleetIcon iconState={iconState} country={meta.country} />
+      {unit.unit.type === "Fleet" && (
+        <WDFleetIcon iconState={unitState} country={unit.country} />
```
TimothyJones, line 51: I like this `unitState` rename. Good catch.

beta-src/src/components/map/components/WDArrowContainer.tsx
```diff
@@ -14,6 +14,7 @@ import ArrowColor from "../../../enums/ArrowColor";
 import drawArrowFunctional, {
   getTargetXYWH,
   getArrowX1Y1X2Y2,
+  makeSVGDrawAsUnsavedAnimateElement,
```
TimothyJones, line 17: This name doesn't feel like it's as clear as it could be. Can it be more about what it is for, not how it does it?

beta-src/src/components/ui/WDGameProgressOverlay.tsx
```diff
@@ -59,7 +59,32 @@ const WDGameProgressOverlay: React.FC<WDGameProgressOverlayProps> = function ({
   if (
     ["Diplomacy", "Retreats", "Builds", "Finished"].includes(overview.phase)
   ) {
-    if (status.phases.length <= 1) {
+    if (status.status === "Left") {
+      innerElem = (
```
TimothyJones, line 63: Would this be better as a separate component?

beta-src/src/components/ui/WDInfoDisplay.tsx
```diff
@@ -50,6 +53,13 @@ const WDInfoDisplay: React.FC<WDInfoDisplayProps> = function ({
     device === Device.MOBILE ||
     device === Device.MOBILE_LG;
 
+  const { phaseMinutes, phaseMinutesRB } = useAppSelector(gameOverview);
+  let phaseLengthInfo = `${getFormattedTime(phaseMinutes * 60)} / phase`;
+  if (phaseMinutesRB !== -1) {
+    phaseLengthInfo = `${phaseLengthInfo} (M) | ${getFormattedTime(
```
TimothyJones, line 59: Could you roll this into the `getFormattedTime` function and avoid the need for the `let`?

In general, there's calculation here that feels like it should belong inside a formatter function rather than inside the render function. You could extract the whole thing to a `getPhaseLengthInfo`

Another aside - I think it's good practice to avoid using the prefix `get` if it's not actually getting anything out of a state / object (for example, `getFormattedTime` implies that the time has already been formatted, not that this function is doing it). For a formatter, I would personally use just `formatTime` / `formatPhaseLengthInfo` etc.

beta-src/src/components/ui/WDInfoDisplay.tsx
```diff
@@ -50,6 +53,13 @@ const WDInfoDisplay: React.FC<WDInfoDisplayProps> = function ({
...
+      phaseMinutesRB * 60,
+    )} / phase (R, B)`;
+  }
   const width = isMobile ? 260 : 320;
```
TimothyJones, line 63: This would be better done in CSS.

beta-src/src/components/ui/WDMessage.tsx
```diff
@@ -33,24 +37,22 @@ const WDMessage: React.FC<WDMessageProps> = function ({
...
+      ? "end"
+      : "start";
   const msgTime = new Date(0);
   msgTime.setUTCSeconds(message.timeSent);
   return (
-    <Box sx={{ display: "flex", justifyContent: justify }}>
-      <Box
+    <div className={`flex justify-${justify}`}>
+      <div
         id={`message-${String(message.timeSent)}`}
```
TimothyJones, line 47: You don't need this `String()` call here. Interpolated strings are already coerced.

beta-src/src/components/ui/WDMessageList.tsx
```diff
@@ -24,27 +23,53 @@ const WDMessageList: React.FC<WDMessageListProps> = function ({
...
-  const messageComponents = filteredMessages.map((message: GameMessage) => (
-    <WDMessage
-      key={`${message.timeSent}:${message.fromCountryID}:${message.toCountryID}:${message.message}`}
-      message={message}
-      userCountry={userCountry}
-      allCountries={allCountries}
-    />
-  ));
+  const messagesByTurn: { [key: number]: React.ReactElement[] } = {};
+  filteredMessages.forEach((message: GameMessage) => {
```
TimothyJones, line 35: I think this would be easier to test and reason about as a `map`, as you're immediately about to iterate over the keys of the temporary object that is being constructed here.

beta-src/src/components/ui/WDMessageList.tsx
```diff
@@ -24,27 +23,53 @@ const WDMessageList: React.FC<WDMessageListProps> = function ({
...
+        message={message}
+        userCountry={userCountry}
+        allCountries={allCountries}
+        viewedPhaseIdx={viewedPhaseState.viewedPhaseIdx}
+      />,
+    );
+  });
+  const messageTurnComponents = Object.entries(messagesByTurn).map(
+    ([turn, msgs]) => {
+      const psy = getPhaseSeasonYear(Number(turn), "Diplomacy");
```
TimothyJones, line 51: I'm surprised this isn't a lint error - `Number()` does some funny coercion things in the case of empty strings (`Number("")` returns `0`) and behaves strangely with strings that aren't decimal (it tries to guess the radix). Some implementations even treat strings with a leading `0` as octal.

It's better practice to use `Number.parseInt(turn, 10)` (with the radix explicit so that it won't mistreat leading zeroes). This will correctly return `NaN` if you pass an empty string.

It won't matter if you're careful to always pass decimal strings without leading zeroes, but if you accidentally pass the wrong thing, tracking down the bug can be hard. Better to always use `Number.parseInt(whateverVariable, 10)`

beta-src/src/components/ui/WDPhaseUI.tsx
```diff
@@ -14,13 +14,22 @@ import {
 import WDCountdownPill from "./WDCountdownPill";
 import WDPillScroller from "./WDPillScroller";
 
+const getCurPhaseMinutes = function (phaseMinutes, phaseMinutesRB, phase) {
```
TimothyJones, line 17: Same comment as earlier about `get` - I think these are selectors or renderers, and would be better without the `get` prefix.

beta-src/src/components/ui/WDPress.tsx
```diff
@@ -179,10 +183,14 @@ const WDPress: React.FC<WDPressProps> = function ({
               id="user-msg"
               label="Send Message"
               variant="outlined"
-              value={userMsg}
+              value={messageStack[countryIDSelected] || ""}
               multiline
               maxRows={4}
-              onChange={(text) => setUserMsg(text.target.value)}
+              onChange={(text) => {
```
TimothyJones, line 189: Same comment about not declaring these inline

beta-src/src/state/game/game-api-slice.ts
```diff
@@ -128,6 +128,7 @@ export const setVoteStatus = createAsyncThunk(
     vote: string;
     voteOn: string;
   }) => {
+    console.log("Submitting vote");
```
TimothyJones, line 131: I don't think this is meant to be in the PR (and probably should be a lint error)

beta-src/src/state/game/game-api-slice.ts
```diff
@@ -143,14 +144,25 @@ export const markMessagesSeen = createAsyncThunk(
     gameID: string;
     seenCountryID: string;
   }) => {
-    const { data } = await getGameApiRequest(
+    const { data } = await postGameApiRequest(
       ApiRoute.MESSAGES_SEEN,
       queryParams,
     );
     return data as string;
```
TimothyJones, line 151: This `as` would be better put inside the `*GameApiRequest` functions, I think. 

Also, since it's just a string that we're looking for, we could check that it's actually a string and throw an error if it isn't, too.

beta-src/src/state/game/game-api-slice.ts
```diff
@@ -143,14 +144,25 @@ export const markMessagesSeen = createAsyncThunk(
...
   },
 );
 
+export const setBackFromLeft = createAsyncThunk(
```
TimothyJones, line 155: I personally would not name these `set` as they can be confused with state setters. For consistency with the other function you could call it `markBackFromLeft`.

I like `update` for a prefix for server updates, and `fetch` for server gets.

In this case I would name the function `markReturnedToGame`

Also, probably it should be a `put` not a `post`, because (unless I'm missing something) this particular request is idempotent.

TimothyJones (commented):
Hi there! I partially reviewed the react code. Nothing show stopping, but I left a number of optional comments about style improvements and risk reduction. If this is useful, I'm happy to review the rest of it.

kestasjk:
Righto didn't get the anonymous game suspicion stuff done, but I've wrapped everything up and started a new release 1.70, and I've tested out a fresh install from the docker composer to check that it's all working. Next dev cycle probably won't be for a few weeks so that should let you make your changes without any moving parts.

beta-src/src/components/controllers/WDMainController.tsx
adamlerer: True, although I believe this component renders rarely (only on phase updae).

beta-src/src/components/controllers/WDMainController.tsx
adamlerer: @lightvector wrote this code, but I believe the reason that this is stateful is that on a phase change, we want to remember the last phase that was being viewed, and let the user manually scroll forward to the current phase. Therefore, it's not a stateless function of the current state.

beta-src/src/components/ui/WDMessageList.tsx
adamlerer: Done. Yeah, the original sin here is that we haven't been disciplined about always treating numbers as numbers - some numbers are retrieved as strings from the server and stored that way, and it will take a careful refactor to fix this everywhere.

beta-src/src/state/game/game-api-slice.ts
adamlerer: I don't know why this `as string` is here - the linter isn't complaining even when I remove it.

adamlerer:
Okay, I've merged in the latest master, things seem to be working correctly now.

@kestasjk there are still a couple things that aren't working quite properly for me right now:
1. The UI behaves weirdly when you cancel a game, in either the old or new UI. To repro: create a bot game, then press cancel (in either UI). 
2. When I run `docker-compose up`, memcached doesn't work. The reason is that the memcached server is exposed as `memcached:11211`, while the PHP code looks for `127.0.0.1:11211`. Solutions I've found are either to change the PHP code to `$MC->addServer("memcached", 11211);` or `docker exec` into the PHP machine and set up memcached there.

@TimothyJones I addressed some of your comments, but some of them would be better suited to a follow-up PR. It would certainly be great if you had time to improve the beta code systematically, there's definitely a lot of things that could be cleaned up and made more idiomatic. But I don't think fixing them locally in just a single place is worth it, and I don't want to do too much more in this PR since it's already quite large.

PR #812, opened 2022-07-13
Fix last phase of finished games for API for beta UI

lightvector:
This PR fixes the API for the last turn of finished games, the handling of which is currently broken in the beta UI for spectacting already-finished games.

PR shipped over from https://github.com/adamlerer/webDiplomacy/pull/138, see there for details and comments.

### DIFF
--- a/api/responses/game_state.php
+++ b/api/responses/game_state.php
@@ -373,6 +373,7 @@ class GameState {
 							"Wait" => "Builds",
 							"Destroy" => "Builds");
 
+		$maxOrderTurn = -1;
 		while( $row = $DB->tabl_hash($orderTabl) )
 		{
 			$order = new \webdiplomacy_api\Order(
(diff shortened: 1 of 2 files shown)


lightvector:
@kestasjk - Would it be better for this PR to be to staging instead of master? Staging appears to be somewhat behind master right now since a lot of the beta UI work bypassed it, but if staging were updated, I can repoint this PR to staging if needed. 

In any case, this PR is to fix a problem that currently that reviewing/spectating the final phase's orders (or even the final season) of finished games doesn't work properly in the new beta UI, due to the database seeming to store a duplicate copy of the last turn's orders for games finished by drawing or concession. I assume this is purposeful, but the API didn't handle it well (it probably hadn't needed to before), so this PR has some work to make the API handle it, plus some unwinding of the beta UI hacks that are no longer needed.

lightvector:
Closing this PR and reopening as a PR to staging.

lightvector:
https://github.com/kestasjk/webDiplomacy/pull/816

PR #824, opened 2022-08-02
Add links for beta spectation

adamlerer:
Even though spectation was possible in the beta UI, there were no links to make it obvious how to do it. Added the links.

image

image


kestasjk:
Oops didn't spot this, thanks for that.

PR #844, opened 2022-09-23
Auto fill support coast

adamlerer:
It was giving errors before, now it fills in the correct coast move based on map click.

https://user-images.githubusercontent.com/5702157/192032774-c41c19f1-0b50-42a6-93cb-deae0527b194.mov


adamlerer:
Oops, this is integrated into https://github.com/kestasjk/webDiplomacy/pull/843

PR #851, opened 2022-10-07
Improvements to center counts in beta UI

adamlerer:


adamlerer:
- Fixes bug where center counts are showing up incorrectly in pre-game
- Fixes bug where supports could auto-fill a move to an invalid coast.

PR #860, opened 2022-11-23
Create cache folder with .htaccess file

alexintour:
without the /cache folder, the variant will throw errors on new installations


kestasjk:
Thanks Alex, not sure how I missed that one

PR #868, opened 2023-06-03
Adding thick borders to the buildable Supply Centers to the Ancient Mediterranean variant via Indexed PNGs

Moxanity:
My previous pull request (#687) didn't use Indexed PNGs resulting in errors. This one is using Indexed PNGs, exported using GIMP.


kestasjk:
Thanks for offering this Moxanity. Is this variant on vDip?

Moxanity:
Yes, it is. Why?
https://www.vdiplomacy.com/variants.php?variantID=9

